PR #56, opened 2017-05-21
Integration Services.

ianamason:
Aloha Bruno, 

This is not ready to merge yet, but I wanted to put it in the queue so that other
people do not waste their time duplicating the effort.  When it is finished we should
have functional travis, coverity, and coverage badges on the README.md.

So far we only have a working travis badge (and it is currently pointing at my fork).

@dddejan  there is a bit of a libpoly mystery that you can see in the build.sh script.
Not sure why libpoly.so.0 is so hard to find.

Mahalo, Ian.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ compiler:
 install:
   - sudo apt-get update
   - sudo apt-get install  -y libgmp-dev
-  - git clone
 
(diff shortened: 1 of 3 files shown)


ianamason:
OK lets see if the coverage stuff is easy.

PR #63, opened 2017-06-13
Not perfect (-dbg needs TLC) but good enough for the debian mentoring process.

ianamason:
# Instructions

clone the repository
```
>git clone https://github.com/ianamason/yices2.git
```
create the upstream tar ball (we should use the github release mechanism for this eventually)
```
>tar cvfz yices2_2.6.0.orig.tar.gz --exclude-vcs yices2
```
create the deb packages:
```
>cd yices2
>debuild -uc -us
```
Note that the version numbers in the upstream tar ball are important.


dddejan:
man pages FTW :)

dddejan:
I was always confused by dependency on gperf for compile. @BrunoDutertre why is this?

debian/control
```diff
@@ -0,0 +1,44 @@
...
+ Long description goes here.
+
+Package: yices2-dev
+Section: libdevel
+Architecture: any
+Depends: yices2,  libyices2.6 (= ${binary:Version}), ${misc:Depends}, libgmp-dev
+Description: C library for manipulating multivariate polynomials
+ Used by the SMT solver yices2.
+ .
+ This package contains development headers for libpoly.
```
dddejan: libpoly -> yices2

ianamason: Right. Though it might actually contain the symbols for libpoly, because I don't think
it contains those for yices yet :-)

ianamason:
The man pages were generated by help2man. No doubt they will be rewritten by BD.

ianamason:
The problem with the -dbg might be that the yices install already does a strip? Could
that be @BrunoDutertre

BrunoDutertre:
We need gperf at compile time to construct perfect hash functions (they are 
used by the lexers).

BrunoDutertre:
The first call to make is the one that strips the libraries and binaries.
Not the 'make install'

ianamason:
OK so this is now on par with the libpoly packing. there is a "STRIP=echo" hack
to prevent the yices2 makefile from stripping, which could be a MODE @BrunoDutertre .
And the man pages are not upto the usual yices2 standard of perfection. But this
is good enough to use in the mentoring process I hope.

BrunoDutertre:
Good enough. 

Fixes to consider in the future:
- remove the "strip hack" and use a MODE=debian or something like that instead
- reformat and expand the man pages

I'll merge the pull request tonight unless Dejan objects.

PR #87, opened 2018-05-18
Mcsat incremental

dddejan:
This is passing basic tests and works with Sally.

@BrunoDutertre please check the changes outside of the mcsat branch:
- changes to context
- changes to model (I now sort output to make output comparable regardless of the engine)


BrunoDutertre:
Merging by hand into master.

PR #99, opened 2018-09-19
Minor correction to clean interrupt

aman-goel:
In clean interrupt mode if the query is UNSAT even before calling yices_check_context, an extra call to smt_pop can happen (due to context_clear_unsat). Since this pop does not have a corresponding push (since start_search hasn't been called), it leads to a segfault in such a case.

The fix is simple. Use a boolean interrupt_push to decide whether to pop or not.


dddejan:
`.directory` files don't seem to belong...

aman-goel:
@dddejan Thanks for the review. Please take a look at the updates.

dddejan:
@aman-goel, do you have a test case that exemplifies the bug?

aman-goel:
Here is a small example. It will produce a seg fault on the master branch.

```
/*
 * YICES API: Test for interactive mode
 * Author: Aman Goel (amangoel@umich.edu)
 * Date: 09/20/2018
 */

#include <assert.h>
#include <stdio.h>

#include <yices.h>

int main(void) {
  ctx_config_t *cfg;
  context_t *ctx;

  printf("Testing Yices %s (%s, %s)\n", yices_version,
  yices_build_arch, yices_build_mode);

  yices_init();
  
  
  cfg = yices_new_config();
  yices_default_config_for_logic(cfg, "QF_UF");
  yices_set_config(cfg, "mode", "interactive");
  ctx = yices_new_context(cfg);
  
  type_t utt = yices_new_uninterpreted_type();

  term_t x = yices_new_uninterpreted_term(utt);
  term_t y = yices_new_uninterpreted_term(utt);
  term_t zero = yices_new_uninterpreted_term(utt);
  
  yices_set_term_name(x, "x");
  yices_set_term_name(y, "y");
  yices_set_term_name(zero, "zero");

  term_t f1 = yices_parse_term("(= x y)");
  term_t f2 = yices_parse_term("(= x zero)");
  term_t f3 = yices_parse_term("(/= y zero)");
  
  yices_assert_formula(ctx, f1);
  yices_assert_formula(ctx, f2);
  yices_assert_formula(ctx, f3);
  
  switch (yices_check_context(ctx, NULL)) {
    case STATUS_SAT:
      printf("The formula is satisfiable\n");
    break;
    case STATUS_UNSAT:
      printf("The formula is not satisfiable\n");
    break;
    case STATUS_UNKNOWN:
    printf("The status is unknown\n");
      break;
    case STATUS_IDLE:
    case STATUS_SEARCHING:
    case STATUS_INTERRUPTED:
    case STATUS_ERROR:
      fprintf(stderr, "Error in check_context\n");
      yices_print_error(stderr);
    break;
  }
  
  yices_free_context(ctx);
  yices_free_config(cfg);
  yices_exit();

  return 0;
}
```

BrunoDutertre:
Thanks for the fix.

PR #115, opened 2019-03-26
Feature var order option

disteph:
Added yices option for fixing variable order in mcsat:  :yices-mcsat-var-order.
--trace mcsat::decide currently shows this order before decisions are made

### DIFF
--- a/src/context/context.c
+++ b/src/context/context.c
@@ -5427,6 +5427,7 @@ void delete_context(context_t *ctx) {
   }
 
   delete_gate_manager(&ctx->gate_manager);
+  /* delete_mcsat_options(&ctx->mcsat_options); // if used then the same memory is freed twice */ 
 
   delete_intern_tbl(&ctx->intern);
   delete_ivector(&ctx->top_eqs);
--- a/src/context/context.c
+++ b/src/context/context.c
@@ -5427,7 +5427,6 @@ void delete_context(context_t *ctx) {
   }
 
   delete_gate_manager(&ctx->gate_manager);
-  /* delete_mcsat_options(&ctx->mcsat_options); // if used then the same memory is freed twice */ 
 
   delete_intern_tbl(&ctx->intern);
   delete_ivector(&ctx->top_eqs);
--- a/src/frontend/common/parameters.c
+++ b/src/frontend/common/parameters.c
@@ -362,6 +362,7 @@ bool param_val_to_factor(const char *name, const param_val_t *v, double *value, 
 bool param_val_to_terms(const char *name, const param_val_t *v, ivector_t **value, char **reason) {
   if (v->tag == PARAM_VAL_TERMS) {
     *value = v->val.terms;
+    return true;
   }
   *reason = "list of variables required";
   return false;
--- a/src/frontend/common/parameters.c
+++ b/src/frontend/common/parameters.c
@@ -71,6 +71,7 @@ static const char * const param_names[NUM_PARAMETERS] = {
   "mcsat-nra-bound-min",
   "mcsat-nra-mgcd",
   "mcsat-nra-nlsat",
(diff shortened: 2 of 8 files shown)


dddejan:
Let's agree on what the option means:

```smt2
(set-logic QF_BV)

(declare-fun x () Bool)
(declare-fun y () Bool)
(declare-fun z () Bool)

(set-option :yices-mcsat-var-order (x y z))

(assert (or x y z))

(set-option :yices-mcsat-var-order (z y x))

(check-sat)
```

If I run yices on above I get:
```
$ ./yices_smt2 --mcsat test.smt2 --trace mcsat::decide
mcsat_decide(): var_order is x y z z y x 
mcsat_decide(): with bool_plugin
mcsat_decide(): variable x
mcsat_decide(): var_order is x y z z y x 
mcsat_decide(): with bool_plugin
mcsat_decide(): variable y
mcsat_decide(): var_order is x y z z y x 
sat
```

Also, if I run it in incremental mode it doesn't seem to work:
```
$ ./yices_smt2 --mcsat test.smt2 --trace mcsat::decide --incremental
mcsat_decide(): with bool_plugin
mcsat_decide(): variable z
mcsat_decide(): with bool_plugin
mcsat_decide(): variable y
sat
```

disteph:
Great! Thanks for this.

Semantics:
- So far I can't see a useful use case for setting the option twice, but I haven't forbidden it and the behaviour of concatenating the sequences seems OK to me; perhaps overriding the previous sequence is more natural, or I could simply forbid to set the option twice.
- Regardless of whether setting the option twice is allowed, should we allow repetitions in the sequence? I haven't forbidden it but only the first occurrence of a variable matters, given the way the decisions uses the option (which is the way I had in mind).
- Allowing empty sequence: Again, I don't see a useful use case for this. At the moment I've put a failing assertion, but I can certainly remove it.

Incrementality:
I didn't write any code to make the option available through the API usage of yices, as the option uses a allocated ivector in the __smtlib2_globals variable. I'll see whether or how that impacts the incremental mode.

disteph:
> Also, if I run it in incremental mode it doesn't seem to work:

Actually, that has nothing to do with the new option, as it seems this was already the case with the other mcsat options: if I set `(set-option :yices-mcsat-nra-bound-max 1)` instead of `(set-option :yices-mcsat-var-order (x y z))`, the value does not survive into the run either if I'm in incremental mode.
That's because in incremental mode, the line `g->ctx->mcsat_options = g->mcsat_options;` of function `init_smt2_context` of file `smt2_commands.c` is executed before the set-options are parsed and treated, whereas in the standard mode the options are parsed and treated first, then that line is executed.
Unless there's a purpose in doing things this way that I'm missing, I can cook up a fix so that all mcsat options that are set survive into the run.

dddejan:
I'll take a look today, but we should involve @BrunoDutertre to figure out the options in general. 

Additionally, options should support 

```
(get-option :yices-mcsat-var-order)
```

And this currently results in a fatal error.

dddejan:
I added get-option support for mcsat options, check out the commit. You can add the same for var-order.

dddejan:
We could do what is done in the ICHECK parameter and explicitly set options in the context if the context is already constructed...

```c
case PARAM_ICHECK:
    if (param_val_to_bool(param, val, &tt, &reason)) {
      g->ctx_parameters.splx_periodic_icheck = tt;
      context = g->ctx;
      if (context != NULL) {
        if (tt) {
          enable_splx_periodic_icheck(context);
        } else {
          disable_splx_periodic_icheck(context);
        }
      }
    }
    break;
```

@BrunoDutertre, would that be the best way to go?

BrunoDutertre:
Yes, I talked with @disteph and that's what he's doing.

PR #116, opened 2019-04-01
Revisit of the eq_ext_conc explainer. Bug fixes

disteph:
I've revisited the eq_ext_conc explainer to fix some bugs and make it more robust / more general. Slicing now only looks at slices of the conflict variable, because everything else can be evaluated. I've rewritten the fragment detection mechanism so that it promises to treat the kind of problems found in the ext-con directory of the bv-problems repo, and aligned the explanation mechanism with the promise.
One bug was fixed in the slicing mechanism itself, and another one was fixed once I realised that the conflict variable could be of type bool and not necessarily bv[n].


dddejan:
In examining some performance regressions I've noticed a conflict being reported in two different forms:

```smt2
(= (bv-extract 1000 1023 a) (bv-concat dummy32 (bv-extract 1008 1023 v7))), (/= (bv-extract 1008 1023 v7) (bv-extract 1008 1023 a))
```

and 

```smt2
(= (bv-extract 1000 1023 a) (bv-concat dummy32 (bv-extract 1008 1023 v7))), (/= (bv-extract 1008 1023 a) (bv-extract 1008 1023 v7))
```

Any ideas why the last disequality has terms in a different order? 

It's not necessarily an issue, as the equalities are normalized by term id's, which can be different if you create terms in different order now. But if there is a simple explanation that you can see, let me know.

disteph:
That looks like a Bruttomesso example...
But no, I can't think of anything smart to say about that order.
Are the two lines you copied: the lemma that I produce for the same conflict, but on two different commits (before / after my PR)? or are they two lemmas that I produce in the same run after my PR?

dddejan:
> Are the two lines you copied: the lemma that I produce for the same conflict, but on two different commits (before / after my PR)? or are they two lemmas that I produce in the same run after my PR?

Two different commits. I was looking at what's different and noticed this. Basically I was trying to see what has changed by looking at the ext_con examples before/after.

dddejan:
Seems like the differences are because some conflicts that were previously described by arithmetic are now handled by equality. This then changes explanations, which in some cases results in worse performance.

disteph:
An easy thing to check is to give precedence to the arithmetic theory in case both of them can explain the conflict. But then again I suspect that'll improve half of the benchmarks and worsen the other half...

PR #122, opened 2019-06-08
Lower bits extraction in bv-arith explainer, plus slight generalisation of ext_conc explainer

disteph:
Compared to the mcsat-branch on official repo, this merge brings in
a) the extension of the bv_arith explainer to explain propagations (not just conflicts)
b) the extensions of the bv_arith explainer to deal with lower bits extraction (both for explaining conflicts and for explaining propagations)
c) a slight extension of the domain of applicability of the extract-concat explainer.

a) and b) are part of the SMT-comp 2019 submission.
c) almost made it but there was a bug in treating brummayerbiere3/maxandminor008.smt2, so it didn't. The bug is now fixed.


dddejan:
Ok, I'm running some checks and will merge later today.

PR #131, opened 2019-07-17
BV implementation of MCSAT

dddejan:
I've checked the following:
- no significant regressions on QF_NIA, QF_NRA, QF_UFNIA, QF_UFNRA
- QF_BV, QF_UFBV doesn't have wrong results or crashes

TODO:

- [ ]  any changes outside of mcsat/ directory should be vetted by @BrunoDutertre to make sure no impact on DPLL(T) engine


BrunoDutertre:
There's also a deadlock when the code is compiled with --enable-thread-safety.

BrunoDutertre:
Correction: there are several deadlocks:

``call_mcsat_solver`` takes the global lock

but within the ``mcsat/bv`` plugin there are several calls to the yices_api
and some of these attempt to get the lock. Need to discuss this with
@ianamason.

ianamason:
I'm all ears.

dddejan:
> Correction: there are several deadlocks:
> 
> `call_mcsat_solver` takes the global lock
> 
> but within the `mcsat/bv` plugin there are several calls to the yices_api
> and some of these attempt to get the lock. Need to discuss this with
> @ianamason.

It seems like there are many calls to the main API, this is the one I found

- [ ] yices_assert_formula
- [ ] yices_bitextract
- [ ] yices_bvconcat
- [ ] yices_check_context
- [ ] yices_check_context_with_assumptions
- [ ] yices_default_config_for_logic
- [ ] yices_delete_term_vector
- [ ] yices_eq
- [ ] yices_free_config
- [ ] yices_free_context
- [ ] yices_get_unsat_core
- [ ] yices_init_term_vector
- [ ] yices_new_config
- [ ] yices_new_context
- [ ] yices_new_uninterpreted_term
- [ ] yices_pop
- [ ] yices_pp
- [ ] yices_print_error
- [ ] yices_push
- [ ] yices_rational
- [ ] yices_term
- [ ] yices_type_of_term

I was thinking that I would just replace them with `_o_` versions. But then I tried with `yices_new_context` and it didn't end well.

Maybe @ianamason  can take a look since it seems like he already did this for nra_plugin. Basically if you run `make check MODE=debug REGRESS_FILTER=mcsat` it will start deadlocking on examples and you can take a look.

ianamason:
I already tried this a while ago. I think I found recursive callbacks or chains of callbacks starting and ending in an attempt to get the lock. Our locks are not recursive so this ends in deadlock. 

I am happy to work with you on this.

dddejan:
@ianamason so I'll try to go through these one by one. 

so if I run it on tests/regress/mcsat/bv/QF_BV_needham.2.prop3_cc_ref_max.smt2 what happens is the following:
- mcsat solving gets the lock
- then the mcsat solver will create a new instance of a regular yices solver
- it will assert things to that solver to check (yices_assert_formula)
- first thing yices_assert_formula does is check the lock

should we have _o_ version of yices_assert_formula?

ianamason:
I am in the SFO cell phone lot. 

I will get back to you tomorrow?

ianamason:
Yes, there is a problem there.  The hope was to make operations on contexts as lock free as possible,
since a context should be isolated on a particular thread. The assert family only grab the lock to
sanity check the operation, so there are not _o_ versions, yet. But if you call them there should be.

Let me think about the best way to do this.

ianamason:
So in a perfect world @dddejan `call_mcsat_solver` would not grab the lock.
We want to allow as much concurrency as possible, so solvers working on separate
contexts should not slow each other down (much).  Accessing the global tables needs
synchronization, but that should be all.  

I tried (without too much vigor) to do this a while back, but all the call backs and perhaps
even recursive calls to the solver itself made me a bit cautious. 

The way I worked was by using helgrind to detect any races or unprotected accesses to the 
global tables.

ianamason:
@dddejan  and @BrunoDutertre  seems like a lot of the races come from calling
unprotected into the term_manager routines.  One solution would be to provide
locking versions of those routines in a systematic way like the _o_ variants of the
yices_api routines, perhaps prefixed by _s_ for syncronized?

dddejan:
@BrunoDutertre i fixed the build and it's synced with master. You can pull the trigger whenever you're ready.

Issue #157, opened 2019-12-09
Assertion failed: (arith_is_zero(terms, ts->garbage)), function bv_arith_coeff


dddejan:
```smt2
(set-logic QF_BV)
(declare-fun x () (_ BitVec 1))
(declare-fun y () (_ BitVec 5))
(declare-fun z () (_ BitVec 5))
(assert (let ((e4 (_ bv4 4))) (let ((e5 (ite (bvule z y) (_ bv1 1) (_ bv0 1)))) (let ((e8 (bvadd ((_ sign_extend 3) x) e4))) (let ((e9 (bvmul e8 ((_ sign_extend 3) e5)))) (let ((e26 (bvslt e9 (_ bv0 4)))) (let ((e52 (=> true e26))) (let ((e55 e52)) (let ((e56 (= e55 true))) (let ((e57 (or e56 e56))) (let ((e71 (xor e57 true))) (let ((e72 e71)) e72))))))))))))
(check-sat)
```

Currently gives 

```console
$ yices_smt2 --mcsat error1.smt2 
Assertion failed: (arith_is_zero(terms, ts->garbage)), function bv_arith_coeff, file mcsat/bv/explain/arith.c, line 286.
```

The backtrace 

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff741f649a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff742b36cb libsystem_pthread.dylib`pthread_kill + 384
    frame #2: 0x00007fff7417ea1c libsystem_c.dylib`abort + 120
    frame #3: 0x00007fff7417dcd6 libsystem_c.dylib`__assert_rtn + 314
    frame #4: 0x000000010028ef43 yices_smt2`bv_arith_coeff(exp=0x0000000100909b20, u=32, assume_fragment=false) at arith.c:286:3
    frame #5: 0x000000010029199d yices_smt2`can_explain_conflict(this=0x0000000100909b20, conflict_core=0x00007ffeefbfebe0, conflict_var=3) at arith.c:1392:24
    frame #6: 0x000000010027af07 yices_smt2`bv_explainer_get_conflict(exp=0x0000000100908170, conflict_in=0x00007ffeefbfebe0, conflict_var=3, conflict_out=0x00007ffeefbfecb8) at bv_explainer.c:193:9
[...]

PR #158, opened 2019-12-10
Fixing issue157.smt2 (over-zealous assert), and adding the regress test

disteph:
Little fix. Passes regress tests, including issue157.smt2.


disteph:
Fixes #157

PR #166, opened 2020-02-08
Allow build using GNU GCC on MacOS

arjenroodselaar:
src/Makefile:644 appends '-Qunused-arguments' to CFLAGS, which assumes Clang is
used for the build. This PR adds some detection to distinguish between
clang-disguised-as-gcc and true GNU GCC, omitting the argument in the latter
case.

This change was tested in the following configurations:
- System compiler (/usr/bin/gcc)
- GCC 9.2 from Brew
- LLVM 9 from Brew


BrunoDutertre:
Thanks

PR #208, opened 2020-05-01
Closes #207. Use `get_ef_model()` when getting values in EF mode.

### LINKED ISSUE #207: SMT-LIBv2 command `(get-value ...)` fails after exists-forall `(check-sat)`
Test case:
====
See [here](https://github.com/boqwxp/yices-bug).

Actual behavior:
====

When using `(get-value ...)` after a `(check-sat)` in a `BV`-logic exists-forall SMT-LIBv2 problem, I get the following error:
```
(error "can't build a model. Call (check-sat) first")
```

Obviously this is wrong, because I just called `(check-sat)`.  Interestingly, `(get-model)` still works (see `yices_out` in the test case repo).

Expected behavior:
====
Yices should print the specified value and not generate an error.  Z3, for example, has no trouble with the exact same input file (see `z3_out` in the test case repo).

boqwxp:
I should have just stepped into a debugger.  The problem is obvious, as is the solution.

### DIFF
--- a/src/frontend/smt2/smt2_commands.c
+++ b/src/frontend/smt2/smt2_commands.c
@@ -4520,7 +4520,10 @@ void smt2_get_value(term_t *a, uint32_t n) {
 
   if (check_logic()) {
     // make sure we have a model
-    mdl = get_model(&__smt2_globals);
+    if (__smt2_globals.efmode)
+      mdl = get_ef_model(&__smt2_globals);
+    else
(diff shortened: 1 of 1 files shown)


BrunoDutertre:
Sorry. I didn't see this pull-request earlier. I fixed it myself.

boqwxp:
No worries.  Glad it's fixed. :)

PR #216, opened 2020-05-09
Closes #210.  Adds support for `timeout` command line argument in exists-forall mode.

### LINKED ISSUE #210: yices-smt2 timeout command line argument is ineffective in exists-forall mode
Test case:
====
See [here](https://github.com/boqwxp/yices-bug/raw/master/yices-bug-2.smt2).

I instrumented `parse_command_line()` in `frontend/yices_smt2.c` and added `printf("setting yices-smt2 timeout to %d\n", v);`.

I ran the instrumented binary like so:
`yices-smt2 -t 10 yices-bug-2.smt2`

Actual behavior:
====
`yices-smt2` prints the following, and does _not_ exit after the specified 10s timeout:

```
user@host:~/github/yices-bug$ yices-smt2 -t 10 yices-bug-2.smt2 
setting yices-smt2 timeout to 10
```

Expected behavior:
====

Yices should abort solving and return `unknown` after the specified timeout.

boqwxp:
It was a bit of a journey following the control flow to the right place, but it turned out to be quite easy to implement this.  You even stubbed out `ef_solver_stop_search()` for me already.  Thanks :)

### DIFF
--- a/src/exists_forall/efsolver.c
+++ b/src/exists_forall/efsolver.c
@@ -187,6 +187,22 @@ void delete_ef_solver(ef_solver_t *solver) {
 }
 
 
+/*
+ * Stop the exists-forall search.
+ */
+void ef_solver_stop_search(ef_solver_t *solver) {
+  context_t *exists_ctx, *forall_ctx;
+
+  exists_ctx = solver->exists_context;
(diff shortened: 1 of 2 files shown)


BrunoDutertre:
Thanks.

Issue #227, opened 2020-06-02
mcsat-assumptions branch: refusal to build interpolant


disteph:
Hey, here's another issue:
```
(set-option :produce-unsat-model-interpolants true)
(set-logic QF_BV)

(declare-fun x1 () (_ BitVec 32))
(declare-fun x2 () (_ BitVec 32))
(declare-fun x3 () (_ BitVec 32))
(declare-fun x4 () (_ BitVec 32))
(declare-fun y!0 () (_ BitVec 32))
(declare-fun y!1 () (_ BitVec 32))
(declare-fun y!2 () (_ BitVec 32))
(declare-fun y!3 () (_ BitVec 32))
(declare-fun forall_placeholder3 () Bool)
(declare-fun forall_placeholder2 () Bool)
(declare-fun forall_placeholder1 () Bool)
(declare-fun forall_placeholder0 () Bool)
(declare-fun y!4 () (_ BitVec 32))
(declare-fun y!5 () (_ BitVec 32))
(declare-fun y!6 () (_ BitVec 32))
(declare-fun y!7 () (_ BitVec 32))
(declare-fun forall_placeholder6 () Bool)
(declare-fun forall_placeholder5 () Bool)
(declare-fun forall_placeholder4 () Bool)

(assert forall_placeholder4)

(assert 
 (or 
  forall_placeholder6 
  (or 
   (or 
    (bvsge 
     (bvadd 
      (bvmul #b00000000000000000000000001011011 y!7) 
      (bvadd 
       (bvmul #b11111111111111111111111110100000 y!5) 
       (bvmul #b00000000000000000000000000111000 y!4))) 
     #b00000000000000000000000000100010) 
    (not 
     (bvsge 
      #b00000000000000000000000000100100 
      (bvadd 
       (bvmul #b11111111111111111111111111100000 y!7) 
       (bvadd 
        (bvmul #b00000000000000000000000000011000 y!6) 
        (bvadd 
         (bvmul #b00000000000000000000000001011000 y!5) 
         (bvmul #b00000000000000000000000000011010 y!4))))))) 
   (not 
    (or 
     (not 
      (bvsge 
       #b11111111111111111111111111100111 
       (bvadd 
        (bvmul #b00000000000000000000000000111111 y!6) 
        (bvadd 
         (bvmul #b11111111111111111111111111111101 y!5) 
         (bvmul #b00000000000000000000000000011101 y!4))))) 
     (not 
      (bvsge 
[...]

PR #230, opened 2020-06-11
Bug fixes

disteph:
- Completed the fix for #227 about interpolant for problems that are unsat before assumptions (should be false_term even if there are no assumptions).
- In the API for `check_context_with_model`, removed the check that the terms should have values in the model; it's ok if they don't: when asking for their values, the default ones will be given and that's ok.
- In the same API function, added a call to `context_clear_unsat(ctx)` if status was previously unsat.
- When Yices clears a SAT status, it should remove the assumptions just like it does when clearing an UNSAT status.


disteph:
I assume I shouldn't be worried about the failed Travis CI?

BrunoDutertre:
The make check fails on travis (timeout on two benchmarks). The same thing
happens to me on the ``mcsat-assumptions``  branch so the timeouts are unrelated
to your pull request. I'll let @dddejan push the button if he's OK with it.

dddejan:
I need to cleanup some stuff, no worries about the CI failures.

dddejan:
@disteph check if my changes work for you. there are many corner cases it turns out.

PR #344, opened 2021-01-06
Add support for cross compiling 32-bit windows version on linux with mingw

nevilad:
Example build order:
```
autoconf
./configure --host=i686-w64-mingw32 CPPFLAGS=-I/home/test/static_gmp/include LDFLAGS=-L/home/test/static_gmp/lib --with-static-gmp=/home/test/static_gmp/lib/libgmp.a --with-static-gmp-include-dir=/home/test/static_gmp/include
make static-dist OPTION=mingw32
```

In makefile newarch is explicitly set to i686-w64-mingw32. Since I've tested on only one mingw version, I don't have data to generalize this code.
In configure.ac, code for checking if libgmp is recent enough is changed. The original version failed this check in case of crosscompiling. The new one succeeds without a real check, since there is no easy way to run the code for windows on linux. This modification affects all cross builds, not only on linux for windows. Since the original code failed, I assume crosscompilation was not used, so no other build will break.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,11 @@ ifneq ($(OPTION),)
   ifeq ($(POSIXOS),linux)
     ifeq ($(OPTION),32bits)
       newarch=$(subst x86_64,i686,$(ARCH))
+    else
+    ifeq ($(OPTION),mingw32)
+      newarch=i686-w64-mingw32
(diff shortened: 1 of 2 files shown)


dddejan:
Thanks!

Issue #352, opened 2021-03-12
Testing mcsat-eval-2 gives error: "generalization failed: nonlinear arithmetic"


disteph:
I'm testing model generalization in QF_NRA (originally solving [an instance of NRA which is in SMTLib](https://clc-gitlab.cs.uiowa.edu:2443/SMT-LIB-benchmarks/NRA/-/blob/master/keymaera/dccs-example-node3896.smt2)).

I'm calling Yices through its API. The sequence of calls can be described in an SMTLib-like syntax as
```
(set-option :produce-unsat-model-interpolants true) 
(set-logic QF_NRA) 
(declare-fun v1 () Real) 
(declare-fun eps () Real) 
(declare-fun t1uscore0dollarskuscore1 () Real) 
(declare-fun A () Real) 
(declare-fun B () Real) 
(declare-fun x1uscore2dollarskuscore1 () Real) 
(declare-fun x2 () Real) 
(declare-fun v1uscore2dollarskuscore1 () Real) 
(declare-fun x1 () Real) 
(declare-fun ts1uscore1 () Real) 
(declare-fun trig1 () Bool) 
(declare-fun name1 () Bool) 
(declare-fun y!1 () Real) 
(assert name1) 
(assert
 (or name1
  (or
   (not
    (>=
     (+ x1uscore2dollarskuscore1 (* (- 1) x2)
      (* t1uscore0dollarskuscore1 v1uscore2dollarskuscore1)
      (* (/ 1 2) (* (* t1uscore0dollarskuscore1 t1uscore0dollarskuscore1) A)))
     0))
   (or (>= (+ (* (- 1) x2) x1 (* (* v1 v1) (/ 1 (* 2 B)))) 0)
    (or (>= (* (- 1) A) 0)
     (or (>= (* (- 1) eps) 0)
      (or (>= (* (- 1) B) 0)
       (or (not (>= v1 0))
        (or (>= (+ (* (- 1) x2) x1) 0)
         (or (not (>= v1uscore2dollarskuscore1 0))
          (or (>= (+ x1uscore2dollarskuscore1 (* (- 1) x2)) 0)
           (or
            (>=
             (+ x1uscore2dollarskuscore1 (* (- 1) x2)
              (* (* v1uscore2dollarskuscore1 v1uscore2dollarskuscore1)
               (/ 1 (* 2 B)))) 0)
            (or
             (>=
              (+ x1uscore2dollarskuscore1 (* (- 1) x2)
[...]

Issue #346, opened 2021-01-19
Wrong interpolant


dddejan:
```smt2
(set-option :produce-unsat-model-interpolants true)
(set-logic QF_NRA)

(declare-fun x () Real)
(declare-fun y () Real)
(declare-fun z () Real)

(assert (< x 1))
(assert (< y 1))
(assert (> x (- 1)))
(assert (> y (- 1)))

(assert
  (or
    (> (+ (* x x) (* y y)) 2)
    (> z 0)
  )
)

(check-sat-assuming-model (z) (0))
(get-unsat-model-interpolant)
```

Gives wrong interpolant due to interval inference. The interpolant should only contain the variable `z`.

Current version gives:

```console
$ ./yices_smt2 issue.smt2 
unsat
(or (< (+ 2 (* -1 (^ x 2)) (* -1 (^ y 2))) 0) (< (* -1 z) 0))
```

Issue #324, opened 2020-10-14
Invalid model for QF_NRA formula


rainoftime:
Hi, for this formula, yices 7139142
givens an invalid model

When feeding the model to yices and cvc4, the solvers decide the formula is unsat.

dddejan:
Related to #43

ahmed-irfan:
This has been fixed. We are following the smtcomp format to represent algebraic numbers.

PR #354, opened 2021-03-18
Interpolation/Generalization for nonlinear arithmetic with MCSAT

dddejan:
Changes:

- Fixed-up solving modulo models and interpolation (#346)
- Support for full computation with algebraic numbers (#324)
- Support for generalization in nonlinear arithmetic (without division, #352)
- Extended the API to allow setting up a model with given values (not terms)
- Producing (model) interpolants needs an option to set in the context `yices_set_config(ctx, "model-interpolation", "true")` or SMT `(set-option :produce-unsat-model-interpolants true)`

TODO:
- Figure out how to best report unsupported term kinds in generalization (#352)

Review needed for:
- Model evaluation
- API changes: https://github.com/SRI-CSL/yices2/blob/4c3c4bf37bbf0ed8bcba01e0998ce8a38ef5e904/src/include/yices.h#L3410-L3448

Regressions pass for both Yices2 and Sally, so the changes should be non-breaking.


ianamason:
This looks like it will need to be propagated into all the yices language bindings, yes?

dddejan:
> This looks like it will need to be propagated into all the yices language bindings, yes?

These are new features so it's not necessary ATM but yes, once it's settled, it would be good to have in the bindings too.

PR #378, opened 2021-09-02
Add new API coverage test

0152la:
This test uses `QF_LIA` instead of the usual `QF_NIA`, therefore I have
removed the check for compilation with`mcsat`. Let me know if this is wrong,
or if the theory should be kept to `QF_NIA` for these tests.

Additional coverage:
api/yices_api.c:yices_arith_geq_atom()
terms/term_utils.c:finite_domain_is_nonpos(), exiting via `return false`
                   term_has_nonpos_finite_domain()
                   arith_term_is_nonpos(), case `ITE_SPECIAL`


BrunoDutertre:
That's fine. No need to check for mcsat in this case.

PR #389, opened 2021-10-30
Makefile.build: avoid ldconfig on linux

trofi:
linux-musl target does not provide ldconfig:
    https://wiki.musl-libc.org/faq.html#Q:-Where-is-%3Ccode%3Eldconfig%3C/code%3E?

Let's create explicit hierarchy of symlinks to avoid dependency on ldconfig's behaviour.


ianamason:
@BrunoDutertre ?

BrunoDutertre:
Looks good to me. It didn't seem to cause any problem on my centos box.

ianamason:
OK lets give it a whirl...  Thanks @trofi

PR #409, opened 2022-09-20
fix for autoconf 2.71

ahmed-irfan:
`AC_PROG_CC_C99` is obselete in the newer version of autoconf (2.70 and newer).
This PR fixes replaces `AC_PROG_CC_C99` by `AC_PROG_CC`. 

https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Obsolete-Macros.html


ianamason:
No one has merged this yet?

Issue #402, opened 2022-08-05
`get-value` on array does not inline function


ekiwi:
I have a satisfiable SMT query and I am interested in the value of an array constant. I understand that yices internally models arrays as functions and I don't mind getting a function back from a `get-value` query. However, currently the answer includes an internal function `@fun_3` which is impossible access unless one is calling `get-model`. It would be great if functions like this could be inlined into the response to a `get-value` query.

Current response:
```
((m@0 @fun_3))
```

Expected response:
```
(function m@0
 (type (Array (_ BitVec 5) (_ BitVec 8)))
 (default #b10000000))
```

Ideal response:
```
((as const (Array (_ BitVec 5) (_ BitVec 8))) #b10000000)
```

I know that `as const` is non-standard, however, I am not sure if there is any clear standard on how responses are formulated and using `as const` would be more similar to how CVC4 and Z3 report the value.

This is the query which can be executed with `yices-smt2`:
```
(set-option :produce-models true)
(set-logic QF_ABV)
(declare-fun m@0 () (Array (_ BitVec 5) (_ BitVec 8)))
(declare-fun m_readValue_data_pipe_0@0 () (_ BitVec 8))
(declare-fun _cycles@0 () Bool)
(declare-fun m_readValue_do_rand_r1@0 () Bool)
(declare-fun out_1@0 () (_ BitVec 8))
(define-fun _resetCount@0 () Bool false)
(declare-fun reset@0 () Bool)
(declare-fun writeAddr@0 () (_ BitVec 5))
(declare-fun readAddr@0 () (_ BitVec 5))
(declare-fun in@0 () (_ BitVec 8))
(declare-fun m_readValue_rand_data@0 () (_ BitVec 8))
(define-fun m.readValue.addr@0 () (_ BitVec 5) readAddr@0)
(define-fun m.readValue.data@0 () (_ BitVec 8) (select m@0 m.readValue.addr@0))
(define-fun _cycles_active@0 () Bool (not (bvuge (ite _cycles@0 (_ bv1 1) (_ bv0 1)) (ite true (_ bv1 1) (_ bv0 1)))))
(define-fun _after_1@0 () Bool (not _cycles_active@0))
(define-fun _GEN_1@0 () Bool true)
(define-fun m_readValue_rand_data.en@0 () Bool m_readValue_do_rand_r1@0)
(define-fun _T@0 () Bool (= readAddr@0 writeAddr@0))
[...]

PR #411, opened 2022-09-27
smtlib-model-format default for yices-smt2

ahmed-irfan:
This PR makes the smtlib-model-format model printing option default for the yices-smt2. 

Updates to: Code + tests + manual


disteph:
@BrunoDutertre:
This was one of the requests in #402: yices-smt2 takes SMTLib2 in, it makes sense to make it gives SMTLib2 out.
It is, however, breaking backward compatibility for scripts that parse the models output by yices-smt2.
Breaking backward compatibility means it should get into Yices 2.7, correct?
I'm OK making master target 2.7 already; if we need bug fixes to 2.6.4 we can have a dedicated branch. Objections?

BrunoDutertre:
No objections from me. It makes sense to increase the version number.

PR #445, opened 2023-05-25
use array const-like model for arrays

ahmed-irfan:
Updates array model printing according to the SMTLIB format.


ahmed-irfan:
DO NOT MERGE

This also changes how UF models are printed

PR #449, opened 2023-07-05
Fix windows binaries exiting with code 24

isanych:
Unfortunately I cannot provide reasonable reproducible example - but we had regular cases when binary sporadically exit with code 24 on big multithreaded projects on windows only, when linux binary working fine on the same project on similar hardware.
Looks like windows code intended to be initialized differently vs posix, but as posix code is working and compatible with windows we have not looked into details too much. With this change our windows binaries are working as expected, without issues seen before.


markpmitchell:
Is `yices_init` being called in multiple threads? If so, how many? Is `yices_exit` being called before `yices_init` is being called again? The Windows code would seem to exit with code 24 if there are no more TLS indices -- and there should be plenty.

wojbas:
The Win32 code is wrong in that the TLS should be initialized once (in yices_init), not per thread (as is the case now in yices_per_thread_init), so consumes a slot with every call to yices_per_thread_init, which we call a lot.

PR #450, opened 2023-07-11
Update html doc

ahmed-irfan:


doc/sphinx/source/api-types.rst
disteph, line 1: Wondering about  "tuple variables are not supported": as far as I can tell, they are. Is that an obsolete description? I'd be curious to see which piece of code still raises that error. I'd guess it's just higher-order variables.

doc/sphinx/source/context-operations.rst
```diff
@@ -935,6 +932,186 @@ distribution.
...
+   - If *ctx*'s current status is :c:enum:`STATUS_UNSAT` then the function does nothing
+     and returns :c:enum:`STATUS_UNSAT`.
+
+   - If *ctx*'s status is :c:enum:`STATUS_IDLE`, :c:enum:`STATUS_SAT`,
+     or :c:enum:`STATUS_UNKNOWN` then the function checks whether
+     *ctx* conjoined with the *n* equalities given by *mdl* and *t* is
+     satisfiable. This is done even if *n* is zero.  The function will
+     then return a code as in :c:func:`yices_check_context`.
+
+   - If *ctx*'status is anything else, the function returns :c:enum:`STATUS_ERROR`.
```
disteph, line 1011: *ctx*'s status

doc/sphinx/source/context-operations.rst
```diff
@@ -111,8 +111,8 @@ If you configure a context for a logic such as ``"QF_NRA"`` or ``"QF_UFNIA"``
 then MCSat will be automatically selected.
 
 The MCSat solver does not currently support as many features as the
-DPLL(T) implementation. In particular, MCSat does not yet support push
-and pop.
+DPLL(T) implementation. In particular, MCSat does not yet support
+computing unsat cores.
```
disteph, line 115: However, unsat cores in MCSat can be computed using c:function:: yices_check_context_with_model and c:function:: yices_get_model_interpolant.

disteph (commented):
Comments are in-lined. Question: does this cover all of the 2.6.4 API extensions, as marked in the yices.h file?

doc/sphinx/source/api-types.rst
ahmed-irfan: > Wondering about "tuple variables are not supported": as far as I can tell, they are. Is that an obsolete description? I'd be curious to see which piece of code still raises that error. I'd guess it's just higher-order variables.

These are the exit codes that were missing. So, I added them. 
However, I think you are right that the CTX_EF_HIGH_ORDER_VARS code is not used. 
I will remove it from the doc and the source code in a separate PR. What do you say?

doc/sphinx/source/context-operations.rst
ahmed-irfan: > However, unsat cores in MCSat can be computed using c:function:: yices_check_context_with_model and c:function:: yices_get_model_interpolant.

Done in cfcd6907

ahmed-irfan:
> Comments are in-lined. Question: does this cover all of the 2.6.4 API extensions, as marked in the yices.h file?

Yes, that covers all the changes in Yices.h.

doc/sphinx/source/api-types.rst
disteph: Agreed. I don't see it being ever raised.

PR #452, opened 2023-08-03
MCSAT support push and pop

bobot:
The code is present and it is selected in tests.


disteph:
That change is already in the #450 merge request.

PR #456, opened 2023-09-22
MC-SAT Thread Safety

markpmitchell:
This pull request contains two components (each in its own commit):

1. Someone had _tried_ to add debug code to detect deadlocks where a thread requests a lock that it already has. However, (a) that code was incorrectly conditionalized, meaning that it was never used, even with YICES_MODE=debug, and (b) the code was incorrect because it failed to initialize the POSIX mutex attribute. Fixed.
2. Running the test suite with YICES_MODE=debug with YICES_MODE=debug revealed a number of places where the MC-SAT code was calling yices_* API functions (which take the global lock) rather than _o_yices_* functions (which do not). Since the MC-SAT code is only reachable via the yices_* API, the global lock is already held at that point; there is no reason to take it again.

With these changes, the test suite passes using --enable-mcsat --enable-thread-safety. Of course, the testsuite doesn't exercise the multi-threaded mode and we have no way to prove that MC-SAT is 100% thread-safe. But, this would seem to be a step forward.

### DIFF
--- a/src/mt/yices_locks_posix.c
+++ b/src/mt/yices_locks_posix.c
@@ -20,30 +20,38 @@
 #include <stdio.h>
 #include <assert.h>
 #include <errno.h>
+#include <string.h>
 
 #include "yices_locks.h"
 
-//I see NDEBUG and DEBUG in the code; which is it?
-#ifdef DEBUG
-static pthread_mutexattr_t mta;
-static pthread_mutexattr_t* mattr = &mta;
-#else
-static pthread_mutexattr_t* mattr = NULL;
-#endif
+static void print_error(const char *caller, const char *syscall, int errnum) {
+  char buffer[64];
 
+  strerror_r(errnum, buffer, sizeof(buffer));
+  fprintf(stderr, "%s failed: %s returned %d: %s\n", caller, syscall,
+	  errnum, buffer);
+}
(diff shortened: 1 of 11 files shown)


src/mt/yices_locks_posix.c
```diff
@@ -20,30 +20,38 @@
...
+}
 
 int32_t create_yices_lock(yices_lock_t* lock){
   int32_t retcode;
-#ifdef DEBUG
-  retcode = pthread_mutexattr_settype(&mta, PTHREAD_MUTEX_ERRORCHECK);
-  if(retcode){
-    fprintf(stderr, "create_yices_lock failed: pthread_mutexattr_settype returned %d\n", retcode);
-  }
+#ifndef NDEBUG
```
ahmed-irfan, line 37: minor style thing:
can we #ifdef DEBUG?

markpmitchell:
No, we cannot — because DEBUG is never defined. :-)

Of course, we could define it somewhere, but NDEBUG is the standard macro used to disable assert. And, most of the Yices code base seems to use “#ifndef NDEBUG” to guard other debug code.

Thank you,

--
Mark Mitchell

ahmed-irfan (approved):
LGTM

PR #457, opened 2023-09-23
Improve thread-safety

markpmitchell:
1. Eliminate `yices_smt2_mt` frontend, incorporating multi-thread test functionality into `yices_smt2`.
2. Provide a thread-safe implementation of timeouts.
3. Avoid allocating very large arrays on the stack. (Noticed because it caused threads, which have smaller stacks, to crash.) Variable-sized allocation on the stack is a bad idea if the allocation size is not bounded.
4. For testing purposes, give thread stacks the same size as the main stack. (Necessary to work around the fact that libcudd uses recursion where it should use iteration in at least one place.)
5. Enhance check.sh to allow passing additional options to tests and to allow running individual tests.
6. Clean up error-handling for POSIX threads API.
7. Fix incorrect order of destruction in `yices_exit`.

With these changes, all regression tests pass with `--enable-mcsat --enable-thread-safety` when running the `yices_smt2` frontend with 8 threads.


disteph:
Looks good, thanks!

PR #460, opened 2023-10-04
fix mingw strerror_r issue

ahmed-irfan:
strerror_r is not available under windows.

This PR fixes this by using strerror (thread safe under windows)


markpmitchell:
Looks good to me.

PR #461, opened 2023-10-04
fix mingw timeout error

ahmed-irfan:
- fix pthread issue for ming

- the_timeout variable is used by the mingw implementation, which is not declared if THREAD_SAFE is not defined. However,  it is needed for mingw. (this was changed in the recent PR about thread safety)

### DIFF
--- a/src/utils/timeout.c
+++ b/src/utils/timeout.c
@@ -76,7 +76,7 @@ typedef struct timeout_s {
   timeout_state_t state;
   timeout_handler_t handler;
   void *param;
-#ifdef THREAD_SAFE
+#if !defined(MINGW) && defined(THREAD_SAFE)
   struct timespec ts;
   pthread_t thread;
   pthread_mutex_t mutex;
@@ -84,14 +84,11 @@ typedef struct timeout_s {
 #endif
 } timeout_t;
 
-#ifndef THREAD_SAFE
-
 /*
  * Global structure common to both implementation.
  */
 static timeout_t the_timeout;
 
-#endif
 
 #ifndef MINGW
 
--- a/src/utils/timeout.c
+++ b/src/utils/timeout.c
@@ -84,10 +84,15 @@ typedef struct timeout_s {
 #endif
 } timeout_t;
 
+
+#if !defined(THREAD_SAFE) || defined(MINGW)
+
 /*
  * Global structure common to both implementation.
  */
 static timeout_t the_timeout;
+
+#endif
 
 
 #ifndef MINGW


markpmitchell:
I missed the MinGW case in this code. Sorry!

I don't think using the_timeout makes sense on MinGW if thread-safe. When not thread-safe, the_timeout works -- but it's inherently wrong to use it when threaded. (We could put it in thread-local storage, but there's no reason that creation of a timeout and use of a timeout have to happen in the same thread.) IIUC, your patch would define the_timeout even when `THREAD_SAFE` is defined, which I think is a mistake. 

So, I think the first hunk of the patch is correct (avoid the pthread stuff on MinGW), but the second hunk shouldn't be necessary. Right now, we don't have a thread-safe version of timeouts for MinGW, so we should add:

```
#if defined(THREAD_SAFE) && defined(MINGW)
#error "No thread-safe implementation of timeouts"
#endif
```

Does that make sense?

ahmed-irfan:
> So, I think the first hunk of the patch is correct (avoid the pthread stuff on MinGW), but the second hunk shouldn't be necessary. Right now, we don't have a thread-safe version of timeouts for MinGW, so we should add:
> 
> ```
> #if defined(THREAD_SAFE) && defined(MINGW)
> #error "No thread-safe implementation of timeouts"
> #endif
> ```
> 
This will make the compilation of yices fail under MinGW when using THREAD_SAFE. However, we have external users who use THREAD_SAFE Yices under windows. I don't know if that should have been allowed at the first place. 
Do you see another way to fix the issue?

markpmitchell:
The answer to "should it have been allowed?" depends on how dogmatic we want to be. No platform was actually thread-safe when using timeouts; the timeouts weren't per-thread. But, since we have Windows users relying on THREAD_SAFE, then, IMO, we should fix thread-safe timeouts to work on Windows. I'm happy to do that, if you like -- though I need to figure out how to get access to a Windows box. 

In the meantime, we could define `the_timeout` if `!defined(THREAD_SAFE) || defined(MINGW)`. We shouldn't define it when `THREAD_SAFE` when POSIX threads because the variable is never used -- and should not be used -- in that case.

ahmed-irfan:
> The answer to "should it have been allowed?" depends on how dogmatic we want to be. No platform was actually thread-safe when using timeouts; the timeouts weren't per-thread. But, since we have Windows users relying on THREAD_SAFE, then, IMO, we should fix thread-safe timeouts to work on Windows. I'm happy to do that, if you like -- though I need to figure out how to get access to a Windows box.
> 
> In the meantime, we could define `the_timeout` if `!defined(THREAD_SAFE) || defined(MINGW)`. We shouldn't define it when `THREAD_SAFE` when POSIX threads because the variable is never used -- and should not be used -- in that case.

I updated the code with your suggestion. Thanks!

I will email you about the access to the windows machine that I am using.

markpmitchell:
Looks good!

PR #465, opened 2023-10-07
windows ci -- checking build and test-api on pull requests

ahmed-irfan:


markpmitchell:
Looks good to me. Why is the final test "make check-api" as opposed to "make check"? Are the regression tests too slow?

ahmed-irfan:
> Looks good to me. Why is the final test "make check-api" as opposed to "make check"? Are the regression tests too slow?

Some tests are failing because of printing mismatch under windows. This needs further investigation and may take come time. In the meantime, I wanted to add the window compilation check in CI.

And yes, the windows CI (including running the tests) is too slow. That's why I didn't enable for each commit push -- it is enabled for pull requests.

markpmitchell:
Makes sense!--Mark MitchellOn Oct 7, 2023, at 9:13 PM, Ahmed ***@***.***> wrote:﻿

Looks good to me. Why is the final test "make check-api" as opposed to "make check"? Are the regression tests too slow?

Some tests are failing because of printing mismatch under windows. This needs further investigation and may take come time. In the meantime, I wanted to add the window compilation check in CI.
And yes, the windows CI (including running the tests) is too slow. That's why I didn't enable for each commit push -- it is enabled for pull requests.

Issue #418, opened 2022-11-10
Cannot build per-thread-state branch with mingw64


isanych:
```
make[2]: Entering directory '/tmp/yices2-per-thread-state/src'
gcc -I. -Iinclude -DPER_THREAD_STATE -I/opt/vcpkg/installed/x64-windows/include -DTHREAD_SAFE -DMINGW -D__USE_MINGW_ANSI_STDIO -DNDEBUG -flto -march=core2 -mtune=corei7 -Wall -Wredundant-decls -Wno-format -O3 -fomit-frame-pointer   -c api/context_config.c -o ../build/x86_64-pc-mingw64-release/obj/api/context_config.o
In file included from C:/tools/msys64/mingw64/include/minwindef.h:163,
                 from C:/tools/msys64/mingw64/include/synchapi.h:10,
                 from ./mt/yices_locks.h:32,
                 from ./utils/object_stores.h:29,
                 from ./terms/bvarith64_buffers.h:48,
                 from ./terms/terms.h:187,
                 from ./context/assumption_stack.h:48,
                 from ./context/context_types.h:31,
                 from ./api/context_config.h:50,
                 from api/context_config.c:26:
include/yices_types.h:83:3: error: expected identifier before '(' token
   83 |   STATUS_INTERRUPTED,
      |   ^~~~~~~~~~~~~~~~~~
make[2]: *** [Makefile:1012: ../build/x86_64-pc-mingw64-release/obj/api/context_config.o] Error 1
```
which caused by conflict with STATUS_INTERRUPTED macro
```
# grep STATUS_INTERRUPTED /mingw64/include/winnt.h
#define STATUS_INTERRUPTED ((DWORD)0xC0000515)
```
I'm renaming STATUS_INTERRUPTED locally, but maybe somebody else would have similar issue

ahmed-irfan:
fixed in 789e1a5b54d78e8c55c804e49c47dd001ff1b895

PR #466, opened 2023-10-09
Rename smt status enum

ahmed-irfan:
Towards fixing #418 

Reason for the change:

STATUS_INTERRUPTED is a defined in windows, and we are using it in the smt_status_t enum. This PR renames the symbols used in the smt_status_t enum.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -4539,23 +4539,23 @@ if (code < 0) {
 }
 
 switch (yices_check_context(ctx, NULL)) {
-case STATUS_SAT:
+case SMT_STATUS_SAT:
   printf("The formula is satisfiable\n");
   ...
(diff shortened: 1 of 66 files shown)


markpmitchell:
At some point, we might consider prefixing all constants exposed as part of the API with YICES_, matching the function names. 

FWIW,

--
Mark Mitchell

ahmed-irfan:
> At some point, we might consider prefixing all constants exposed as part of the API with YICES_, matching the function names. FWIW,
> […](#)

+1
That's a good point.

disteph (approved):
Looks good to me. That should go in the next non-backward compatible release though.

ahmed-irfan:
superseded by https://github.com/SRI-CSL/yices2/pull/587

PR #467, opened 2023-10-10
hack: undef status_interrupted to fix windows build

ahmed-irfan:
this hack is a temporary fix for the windows build. 

A proper fix (https://github.com/SRI-CSL/yices2/pull/466) will go into the Yices 2.8 release.

### DIFF
--- a/src/api/yices_api.c
+++ b/src/api/yices_api.c
@@ -8510,7 +8510,7 @@ EXPORTED int32_t yices_push(context_t *ctx) {
     }
     assert(context_status(ctx) == STATUS_UNSAT);
     // fall through
-  case STATUS_INTERRUPTED:
+  case YICES_STATUS_INTERRUPTED:
   case STATUS_SEARCHING:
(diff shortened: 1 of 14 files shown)


ahmed-irfan:
> There’s no backwards compatibility problem on MINGW if the code never compiled before. Did it? If it didn’t, you could do this: 1. In the enum definition use YICES_STATUS_INTERRUPTED. 2. Elsewhere in yices.h add: /* To avoid breaking backwards compatibility until Yices 2.8. */ #ifndef MINGW #define STATUS_INTERRUPTED YICES_STATUS_INTERRUPTED #endif Then, change all the .c files to use YICES_STATUS_INTERRUPTED. That avoids changing the API for 2.7 update releases on non-Windows platforms. And it avoids undef’ing a Windows API macro. FWIW,
> […](#)

I like it. Updating the PR accordingly. 
Thanks Mark!

disteph:
I agree with that solution.

markpmitchell (approved):
Looks good!

PR #468, opened 2023-10-10
Update ci -- run mcsat with thread-safety

ahmed-irfan:
This PR enables running yices-mcsat with the thread-safety option. 
We run it only with GCC, and not with clang because of the unavailability of thread-local storage.


markpmitchell:
Looks good! I was meaning to get to that. Thanks!

PR #471, opened 2023-10-17
Win timeout

### LINKED ISSUE #462: fix thread-safe timeouts to work on Windows
#461 is a workaround for the issue, but a proper solution need to implemented for Windows

markpmitchell:
Avoid use of global variables in timeouts on MinGW to avoid thread-safety issues.

Fixes #462


ahmed-irfan (approved):
LGTM

PR #472, opened 2023-11-01
Mcsat check model with hint API method

ahmed-irfan:
- Adds the new api method
- Adds a test to check the api method
- Updates the documentation

### DIFF
--- a/src/mcsat/solver.c
+++ b/src/mcsat/solver.c
@@ -2522,6 +2522,42 @@ void mcsat_check_model(mcsat_solver_t* mcsat, bool assert) {
 static
 void mcsat_assert_formulas_internal(mcsat_solver_t* mcsat, uint32_t n, const term_t *f, bool preprocess);
 
+void mcsat_set_model_hint(mcsat_solver_t* mcsat, model_t* mdl, uint32_t n_mdl_filter,
+			  const term_t mdl_filter[]) {
+  if (n_mdl_filter == 0) {
+    return;
+  }
+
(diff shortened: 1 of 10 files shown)


src/mcsat/solver.c
```diff
@@ -2522,6 +2522,42 @@ void mcsat_check_model(mcsat_solver_t* mcsat, bool assert) {
...
+    uint32_t plugin_i = mcsat->decision_makers[variable_db_get_type_kind(mcsat->var_db, x_var)];
+    value_t x_value = model_get_term_value(mdl, x);
+    mcsat_value_t value;
+    mcsat_value_construct_from_value(&value, vtbl, x_value);
+    assert(x_value >= 0);
+    trail_add_propagation(mcsat->trail, x_var, &value, plugin_i, mcsat->trail->decision_level);
+    mcsat_value_destruct(&value);
+  }
+
+  mcsat_pop(mcsat);
```
disteph, line 2558: Does the use of push/pop mechanism mean that we lose all of the lemmas we have learnt prior to the call to yices_check_context_with_model_and_hint ?

ahmed-irfan: No, we will not lose the lemmas learned before calling the yices_check_context_with_model_and_hint method. The mcsat_pop method removes those lemmas that were learned after the last call to mcsat_push. Here we don't remove any lemmas because we didn't learn any between the push/pop call.

disteph: OK, awesome.

PR #478, opened 2023-11-20
Mcsat api var order

ahmed-irfan:
This PR :
- adds a method to set a variable ordering for making decisions in the MCSat search.
- moves the variable ordering from the mcsat options to the solver context.
- adds a api test case


doc/sphinx/source/context-operations.rst
```diff
@@ -1207,6 +1207,39 @@ as follows::
      -- error code: :c:enum:`CTX_INVALID_OPERATION`
 
 
+Set Variable Ordering for MCSat
+-------------------------------
+
+It is possible to give a variable ordering for the MCSat search --
+this will make MCSAT to decide the variables in the given order. Note
```
disteph, line 1214: "...make MCSAT decide the variables..."
Can we specify in the doc whether the ordering is forced during the initial search phase only, or if it is persistent across backtracks (i.e. variable activities will not affect the order)?

disteph (approved):

doc/sphinx/source/context-operations.rst
ahmed-irfan: good point. I will add that info

PR #488, opened 2024-02-16
Check parallel

Ovascos:
make check is executing tests in parallel (when either GNU parallel or moreutils' parallel is available)


Ovascos:
Note that this PR is tested with Mac and Linux only. I don't have a Windows+cygwin setup available.

ahmed-irfan:
How do you specify the number of cores to use?

Ovascos:
You can not. All cores are used.

tests/regress/run_test.sh
ahmed-irfan, line 1: is this code essentially coming from the check.sh? If yes, are there any changes?

ahmed-irfan:
> You can not. All cores are used.

Is there an easy way to pass this number?

Ovascos:
> > You can not. All cores are used.
> 
> Is there an easy way to pass this number?

I'll have a look.

Ovascos:
Integrated check.sh within GNU make's token system to use limit the number of total parallel execution to the value specified by -j N

ahmed-irfan (approved):
LGTM

Thanks

PR #494, opened 2024-03-11
Decision hint queue

Ovascos:
Adds a queue to enable plugins to suggest a variable selection to the mcsat core.
Fixed typo in function name "mcsat_process_registeration_queue"
Updates the int_queue


src/mcsat/solver.c
```diff
@@ -2374,6 +2418,7 @@ bool mcsat_decide(mcsat_solver_t* mcsat) {
           var = variable_null;
         } else {
           // fprintf(stderr, "random\n");
+          var_queue_remove(&mcsat->var_queue, var);
```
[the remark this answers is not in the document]
ahmed-irfan: Your changes are correct, but the old code was handling this lazily by checking if a variable has a value in the trail or not. (https://github.com/SRI-CSL/yices2/pull/494/files#diff-9beddeebb3753d511c9f74a6cac286af71b99af654e1bbac026f53de0eefcd47L2389).

src/mcsat/solver.c
```diff
@@ -2703,6 +2748,9 @@ void mcsat_solve(mcsat_solver_t* mcsat, const param_t *params, model_t* mdl, uin
       break;
     }
 
+    // remove all the hints
+    int_queue_reset(&mcsat->hinted_decision_vars);
```
ahmed-irfan, line 2752: I think it is better to move this call in the mcsat_pop_internal method. That way we will also reset the queue when we do a restart.

ahmed-irfan (commented):
One comment needs attention.

src/mcsat/solver.c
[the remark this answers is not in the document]
Ovascos: I will provide a more elaborate fix that takes this into account and avoids any var queue manipulation.

src/mcsat/solver.c
```diff
@@ -2339,32 +2342,88 @@ bool mcsat_decide_assumption(mcsat_solver_t* mcsat, model_t* mdl, uint32_t n_ass
...
     var = variable_null;
+    aux_choice = true;
 
     // Us the top variables first
-    for (i = 0; i < mcsat->top_decision_vars.size; ++i) {
+    for (uint32_t i = 0; i < mcsat->top_decision_vars.size; ++i) {
       var = mcsat->top_decision_vars.data[i];
       assert(var != variable_null);
       if (!trail_has_value(mcsat->trail, var)) {
-        force_decision = true;
```
ahmed-irfan: why remove this?

src/mcsat/solver.c
```diff
@@ -2376,7 +2435,6 @@ bool mcsat_decide(mcsat_solver_t* mcsat) {
         var = int_queue_pop(&mcsat->hinted_decision_vars);
         assert(var != variable_null);
         if (!trail_has_value(mcsat->trail, var)) {
-          force_decision = true;
```
ahmed-irfan: why remove this?

src/mcsat/solver.c
Ovascos: Because it was only a quickfix to avoid getting skipped variables to be re-inserted in the var queue.
Not forcing is the preferred option, as it gives the plugins more flexibility.

ahmed-irfan: I think these variables need to be forced because they are top vars or in the var order given by the user.

src/mcsat/solver.c
ahmed-irfan: this is correct. I understand now. But the earlier one should be changed.

ahmed-irfan (changes requested):
force decision for top-vars.

ahmed-irfan (approved):

PR #520, opened 2024-06-18
Fixes missing comparison in eq_pprod_hobj

### LINKED ISSUE #519: `pprod_term` may return incorrect pprod on hash collision
Function `eq_pprod_hobj` returns true for any pprod, as the returns statement is missing a comparison with the expected result `o->r`: https://github.com/SRI-CSL/yices2/blob/f06761440620cb803af6275baaefd3cbe880bb70/src/terms/terms.c#L1047

Ovascos:
fixes #519


ahmed-irfan (approved):
Thanks for investigating the issue and fixing it.

PR #527, opened 2024-09-03
add timeout keyword to set-option in the smt2 frontend

ahmed-irfan:


nafur:
I checked this on an example, thank you so much for this!

PR #548, opened 2025-01-14
fix return type of unsafe_literal_is_unassigned

isanych:
most likely copy/paste error


ahmed-irfan (approved):
Thanks for fixing this.

PR #564, opened 2025-06-13
yices2_parallel.py

ahmed-irfan:


ahmed-irfan:
add documentation about this in the manual and README

PR #590, opened 2025-07-12
fixes #414

### LINKED ISSUE #414: yices_new_uninterpreted_term does not produce an uninterpreted term
There's a corner case I've just discovered:
If the type is the unit type (scalar type of cardinality 1), yices_new_uninterpreted_term returns the unique inhabitant of the type.
It's satisfiability-preserving, but not ideal.
In particular, yices_subst_term will then complain if you try to substitute what you thought was an uninterpreted term (but is actually this constant) by an expression of type unit.  In a model, I suspect the value for what the user thought was an uninterpreted term may also be missing, because the uninterpreted term never existed in the first place (to be checked).
Obviously, there's little point introducing uninterpreted terms of type unit, perhaps even less substituting them, but the user code may be generic over Yices types, so now we're pushing onto the user the burden to do a case analysis on the type, for an optimization that isn't documented.

@BrunoDutertre : Did the trick provide speed-ups? If we want to keep it, we may want to do a case analysis in yices_subst_term so that it doesn't complain but instead ignores the substitution on type unit, trying to pretend there is actually an uninterpreted term of type unit even if internally there isn't. And perhaps make sure a value is returned in the model for an uninterpreted term of type unit? Any thoughts?

ahmed-irfan:
fixes #414


disteph (commented):
I moved the trick of replacing uninterpreted symbols of unit types by the unique value of the type, out of the term-construction function and into the CDCLT internalization process. This is necessary over not doing the trick at all, because of function types that are unit (the CDCLT e-graph doesn't seem to know that you can't have two distinct terms in that type). However, MCSAT no longer benefits from the trick. But MCSAT has a larger problem of not being aware of cardinality constraints on finite function types. It happily says SAT on 5 pairwise distinct functions of type Bool -> Bool. So I made MCSAT refuse to assert formulas that contain equalities on finite functional types.

disteph:
# MCSAT Function-Disequality Findings

This note records the experiments run with the PR590 unsupported-theory guard temporarily removed.

The tests were run on `fix-iss-414` at commit `2d0d0be9` (`Move unsupported MCSAT regressions out of both`), with the then-current PR590 finite-function guard in `src/mcsat/preprocessor.c` temporarily removed and `yices_smt2` relinked against that no-guard build.

The same four tests were rerun on `origin/master` at commit `485acafd` (`Fix #613 MCSAT curried function model reconstruction (#614)`) after rebuilding and relinking `yices_smt2`. The results are the same: the bug is present on master, not introduced by PR590.

## Finite Function Sort: `Bool -> Bool`

The explicit pairwise-disequality encoding is semantically unsatisfiable: there are only four total functions of type `Bool -> Bool`, so five pairwise-distinct functions cannot exist.

```smt2
(set-logic ALL)

(declare-fun f0 (Bool) Bool)
(declare-fun f1 (Bool) Bool)
(declare-fun f2 (Bool) Bool)
(declare-fun f3 (Bool) Bool)
(declare-fun f4 (Bool) Bool)

(assert (not (= f0 f1)))
(assert (not (= f0 f2)))
(assert (not (= f0 f3)))
(assert (not (= f0 f4)))
(assert (not (= f1 f2)))
(assert (not (= f1 f3)))
(assert (not (= f1 f4)))
(assert (not (= f2 f3)))
(assert (not (= f2 f4)))
(assert (not (= f3 f4)))

(check-sat)
(get-model)
```

With the guard removed, MCSAT returns `sat`. The produced model violates the asserted disequalities; for example, `f0` and `f4` are both the constant-false function:

```text
sat
((define-fun f0 ((x!0 Bool)) Bool false)
 (define-fun f1 ((x!0 Bool)) Bool true)
 (define-fun f2 ((x!0 Bool)) Bool (ite (= x!0 true) true false))
 (define-fun f3 ((x!0 Bool)) Bool (ite (= x!0 false) true false))
 (define-fun f4 ((x!0 Bool)) Bool false))
```

The corresponding `distinct` spelling does not expose the same search bug:

```smt2
(assert (distinct f0 f1 f2 f3 f4))
```

This is not because MCSAT search handles the cardinality constraint soundly. It is because `mk_distinct` simplifies the term before MCSAT search: if `type_card(tau) < n` and the cardinality is exact, it returns `false_term`. For `Bool -> Bool`, the exact cardinality is 4, so `(distinct f0 f1 f2 f3 f4)` is simplified to `false`.

## Finite Domain, Infinite Codomain: `Bool -> Int`

The finite-function guard is not broad enough. A function sort with finite domain and infinite codomain can still have constrained finite image, and then function disequalities require extensional/cardinality reasoning.

This formula is semantically unsatisfiable: every `fi` maps both Boolean inputs into `{0,1}`, so there are only four possible constrained function graphs, but five functions are required to be pairwise distinct.

```smt2
(set-logic ALL)

(declare-fun f0 (Bool) Int)
(declare-fun f1 (Bool) Int)
(declare-fun f2 (Bool) Int)
(declare-fun f3 (Bool) Int)
(declare-fun f4 (Bool) Int)

(assert (not (= f0 f1)))
(assert (not (= f0 f2)))
(assert (not (= f0 f3)))
(assert (not (= f0 f4)))
(assert (not (= f1 f2)))
(assert (not (= f1 f3)))
(assert (not (= f1 f4)))
(assert (not (= f2 f3)))
(assert (not (= f2 f4)))
(assert (not (= f3 f4)))

(assert (or (= (f0 true) 0) (= (f0 true) 1)))
(assert (or (= (f0 false) 0) (= (f0 false) 1)))
(assert (or (= (f1 true) 0) (= (f1 true) 1)))
(assert (or (= (f1 false) 0) (= (f1 false) 1)))
(assert (or (= (f2 true) 0) (= (f2 true) 1)))
(assert (or (= (f2 false) 0) (= (f2 false) 1)))
(assert (or (= (f3 true) 0) (= (f3 true) 1)))
(assert (or (= (f3 false) 0) (= (f3 false) 1)))
(assert (or (= (f4 true) 0) (= (f4 true) 1)))
(assert (or (= (f4 false) 0) (= (f4 false) 1)))

(check-sat)
(get-model)
```

With the guard removed, MCSAT returns `sat`. The model uses unreachable else-branch values to make the printed functions look distinct:

```text
sat
((define-fun f0 ((x!0 Bool)) Int (ite (= x!0 true) 0 (ite (= x!0 false) 0 4)))
 (define-fun f1 ((x!0 Bool)) Int (ite (= x!0 true) 0 (ite (= x!0 false) 0 2)))
 (define-fun f2 ((x!0 Bool)) Int (ite (= x!0 true) 0 (ite (= x!0 false) 0 3)))
 (define-fun f3 ((x!0 Bool)) Int 0)
 (define-fun f4 ((x!0 Bool)) Int (ite (= x!0 true) 0 (ite (= x!0 false) 0 1))))
```

Since `x!0` has type `Bool`, the else values are unreachable. Extensionally, all five functions map both `true` and `false` to `0`.

The same wrong-`sat` behavior occurs with the `distinct` spelling for `Bool -> Int`. Unlike `Bool -> Bool`, the function sort `Bool -> Int` is infinite, so `mk_distinct` cannot simplify `(distinct f0 f1 f2 f3 f4)` by exact type cardinality.

## Consequence for PR590

The temporary unsupported-theory guard should not be limited to finite function sorts. It should reject equality/disequality atoms whose type contains a function sort that needs disequality monitoring:

- function sort with finite domain
- function sort with unit range
- any enclosing function, tuple, or type-instance sort that contains such a function sort by downward closure

The guard is still a temporary solution. The complete fix is to make MCSAT's UF reasoning enforce these extensional/cardinality constraints during search rather than rejecting the input.

disteph (commented):
Review provided by Codex/GPT5.5 High.
I concur.

**Findings**
No blocking findings.

I did not find a correctness issue in the branch diff against `origin/master` at `485acafd`. The MCSAT guard now matches the observed unsafe cases: finite-domain function sorts, unit-codomain function sorts, and downward closure through function domains/ranges, tuples, and type instances. The regression harness change for solver-specific `.mcsat.gold` / `.dpllt.gold` files is scoped to `tests/regress/both` and preserves shared `.gold` behavior by default.

**Open Questions**
The main residual risk is intentional, not a defect: PR590 remains an over-approximating temporary guard. It rejects some formulas that could be simplified or solved soundly, for example benign equalities over guarded function sorts. That matches the stated goal of refusing inputs that require missing MCSAT function-disequality monitoring.

**Review Summary**
The unit-type change in `src/context/context.c` compensates for making unit-typed uninterpreted terms fresh again in `src/terms/term_manager.c`, preserving CDCL(T) singleton semantics during internalization.

The MCSAT guard in `src/mcsat/preprocessor.c` correctly applies to both `EQ_TERM` and `DISTINCT_TERM`, and the new SMT2 regressions cover the exact master wrong-`sat` reproducers under both solver modes.

**Tests Run**
`git diff --check origin/master...HEAD`

`tests/regress/check.sh tests/regress/both build/arm-apple-darwin23.6.0-debug/bin`
Result: `Pass: 11`, `Fail: 0`

`build/arm-apple-darwin23.6.0-debug/bin/issue_414_mcsat`
Result: pass

`build/arm-apple-darwin23.6.0-debug/bin/issue_414`
Result: pass

disteph (approved):

disteph:
Note: "I made MCSAT refuse to assert formulas that contain equalities on finite functional types" -> I generalized the guard since (i) the search was also incorrect when having equalities or disequalities on function types with finite domains and infinite co-domains (see example Bool -> Int above); equalities/disequalities on D->C implicitly generate equalities/disequalities on C and D (so there a downward closure of the guard I added too).

disteph:
Code review by Windsurf/Opus4.7 high

# Code review: PR 590 (master..fix-iss-414, 13 commits, 38 files, +717/-103)

## Scope

Fixes issues #414 and #613 along five axes:

- **Unit-type term creation** — drop the [mk_uterm](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/term_manager.c:3995:0-4001:1) collapse at `@/Users/sgl/git/yices/yices2/src/terms/term_manager.c:3998-4001`.
- **CDCL(T) internalization** — re-add the collapse at the egraph layer in [internalize_to_eterm](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context.c:2550:0-2809:1) (`@/Users/sgl/git/yices/yices2/src/context/context.c:2594-2614`), with consistency asserts at both mapping sites.
- **MCSAT theory guard** — new [term_needs_function_diseq_guard](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:214:0-216:1) rejects `EQ_TERM` and `DISTINCT_TERM` whose type transitively contains a finite-domain or unit-codomain function sort (`@/Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:170-219,553,1153-1160`).
- **MCSAT UF plugin** — curried function model reconstruction plus fail-closed hardening (`@/Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c`).
- **Test infrastructure** — `tests/regress/both/` with `.mcsat.gold` / `.dpllt.gold` overrides, SMT2 and API regressions, documented `.smt2.disabled` convention (`@/Users/sgl/git/yices/yices2/tests/regress/run_test.sh:194-225`, `@/Users/sgl/git/yices/yices2/tests/regress/mcsat/README.md`).

The [vtbl_copy_update_maps](cci:1://file:///Users/sgl/git/yices/yices2/src/model/concrete_values.c:1119:0-1142:1) extraction into `@/Users/sgl/git/yices/yices2/src/model/concrete_values.c` preserves the issue #613 hset1 reentrancy fix across all five call sites.

## Findings

### Medium

**M1 — Self-referential cycles may emit non-extensional models.** `@/Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:875-891` caches `find->val = f_value` before recursing, to break cycles where an application's argument or result shares the head's `function_id`. The cached value at cycle time is the pre-update base, so the emitted model can fail extensionality on the self-referential pair. The KNOWN LIMITATION comment documents this, but no regression test exercises the path. Follow-up: file an issue linked from the comment. A real fix needs a visited-set plus a theory-conflict path.

**M2 — Unbounded recursion in [type_needs_function_diseq_guard](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:172:0-212:1).** `@/Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:170-216` walks `FUNCTION_TYPE`, `TUPLE_TYPE`, and `INSTANCE_TYPE` with no memoization. `tests/regress/both/wd_issue301.smt2` runs a 10+-deep nest fine, but on DAG-shaped types with heavy fan-in the repeat work can blow up. Follow-up: either confirm in a comment that `type_table_t`'s hash-consing amortizes repeat visits, or cache results in an `int_hmap_t` keyed on `type_t` per [preprocessor_apply](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:408:0-1228:1) call.

### Low

**L1 — DISTINCT guard re-checks every argument.** `@/Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:1153-1160` loops over all `n` args, but SMT-LIB requires `distinct` to be homogeneous. Checking `desc->arg[0]` alone suffices and matters on large distincts (LDV traces have hundreds). One-line cleanup.

**L2 — Unit-codomain function equality is trivially decidable but rejected.** `@/Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:175` rejects `f = g : T -> unit` even though all such functions are extensionally equal. Sound but unnecessarily conservative. Flag in the guard comment as deliberately coarse.

**L3 — Consistency asserts in [internalize_to_eterm](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context.c:2550:0-2809:1) are debug-only.** `@/Users/sgl/git/yices/yices2/src/context/context.c:2603-2611,2794-2802` asserts existing root mappings agree with the egraph occurrence being returned. Release builds with `NDEBUG` compile these out. Consistent with the rest of [context.c](cci:7://file:///Users/sgl/git/yices/yices2/src/context/context.c:0:0-0:0); upgrade to a runtime check or `longjmp` only if the project's style dictates.

**L4 — H3 cycle path has no regression.** The known-limitation branch in `@/Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:875` is hard but not impossible to provoke. A smoke test with a heavily updated curried function sharing sub-results would lock in the early-cache-write behavior.

### Info

**I1 — Track degraded MCSAT answers alongside disabled tests.** `@/Users/sgl/git/yices/yices2/tests/regress/both/wd_issue301.smt2.mcsat.gold` is `(error "mcsat: unsupported theory")` while DPLLT returns `sat`. Consider extending `@/Users/sgl/git/yices/yices2/tests/regress/mcsat/README.md` with a sub-table for tests that still pass but with a degraded MCSAT gold.

**I2 — API tests are thorough.** `@/Users/sgl/git/yices/yices2/tests/api/issue_414_mcsat.c` systematically probes the guard boundary across nine cases, including the still-accepted `Int -> Bool` baseline (case 9). `@/Users/sgl/git/yices/yices2/tests/api/mcsat_curried_function_model.c` locks in the issue #613 model path with `YVAL_FUNCTION` assertions on the partial application.

## Verification

```text
git diff --check                                      # clean
make MODE=debug -j8                                   # clean, no -Wall warnings
tests/regress/check.sh tests/regress/both ...         # 11/11 PASS (mcsat + dpllt)
tests/regress/check.sh tests/regress/mcsat/arrays ... # 23/23 PASS (incl. issue613_*)
make MODE=debug check-api                             # 30/30 PASS (incl. issue_414*, mcsat_curried_function_model)
```

Pre-existing `set-timeout.smt2` / `wd/issue401b.smt2` failures are macOS-timeout infrastructure, unrelated.

## Note on the disabled tests

I initially flagged the guard's rejection of `(Array (BV N) T)` equalities as a coverage regression worth tightening. That framing was wrong: the two disabled LDV benchmarks (`linux-...locktorture...`, `32_1_cilled_...gpio_keys_polled...`) each assert direct `(= a (store b k v))` atoms over `(Array (BV 64) T)`. Those equalities are exactly the extensionality cases MCSAT cannot soundly decide without the Plan A/B extensionality work. Pre-PR MCSAT did return the expected gold answers on these inputs, but only because the rest of the Boolean skeleton was consistent with those answers regardless of how MCSAT interpreted the array equalities. Any cardinality-threshold relaxation of the guard would be a soundness regression. The disabled state is the correct one; the re-enable condition is extensional-function support, exactly as the disabled-tests README records.

## Recommendation

**Approve.** The four review-driven hardening commits (`bb24f567`, `8d4154b3`, `a8994060`, `6ac4414a`) close all high-severity findings from the prior round. The remaining medium-severity items are follow-up work, not blockers: M1 is a documented limitation with no current regression; M2 is a performance concern only. The low-severity items are style and minor cleanup.

disteph:
Code review by Codex/GPT5.5

**Findings**

No findings. I reviewed the current `fix-iss-414` branch against `origin/master` at merge-base `485acafd9670`.

**Summary**

The PR now looks mergeable from my review pass. The unit-type uninterpreted-term change is covered by CDCL(T) internalization canonicalization and API regressions. The MCSAT guard now covers both equality and `distinct`, including the finite-domain/infinite-codomain cases and downward closure. The MCSAT UF model hardening avoids the previous release-build UB/collapsing-id paths, and the shared update-map printer helper is a clean extraction.

Residual risk is intentional scope: four MCSAT regressions are disabled behind the temporary finite-function guard, but they are documented in `tests/regress/mcsat/README.md`.

**Tests Run**

- `git diff --check origin/master...HEAD`
- `make MODE=debug -j4`
- `tests/regress/check.sh tests/regress/both build/arm-apple-darwin23.6.0-debug/bin`
- `tests/regress/check.sh tests/regress/mcsat/arrays build/arm-apple-darwin23.6.0-debug/bin`
- `build/arm-apple-darwin23.6.0-debug/bin/issue_414`
- `build/arm-apple-darwin23.6.0-debug/bin/issue_414_mcsat`
- `build/arm-apple-darwin23.6.0-debug/bin/mcsat_curried_function_model`

PR #606, opened 2026-03-02
MCSAT: add support for tuples (incl. nested with function types); blasts tuples in input constraints + reconstitutes them in models and interpolants

disteph:
This branch adds end-to-end tuple support in MCSAT by eliminating tuples before solving and reconstructing results in terms of the original problem.

It includes:
1. Tuple/function preprocessing (tuple blasting) before MCSAT solving.
2. Model reconstruction from blasted leaves back to original tuple-typed terms/functions.
3. Interpolant post-processing to replace tuple leaves with accessor-based terms over original tuple variables.
4. Scope/GC integration so preprocessing artifacts are backtrack-safe.
5. Expanded MCSAT-only API regression tests.

### Preprocessor: tuple blasting
- Extended MCSAT preprocessor state and traversal to blast tuple structure recursively.
- Handles:
  - tuple constructors/selectors
  - tuple equality/distinct
  - tuple-typed ITE
  - function applications/updates involving tuple domain/codomain
  - function type flattening for tuple arguments/results

### Model reconstruction
- Rebuilds tuple atom values from leaf values.
- Rebuilds function values of original tuple types by merging blasted leaf functions:
  - reconstruct tuple-domain arguments from flattened arguments
  - reconstruct tuple codomain outputs from leaf outputs
  - preserve defaults/mappings coherently

### Interpolants
- Added unblasting of interpolants returned by MCSAT:
  - tuple leaves are rewritten to selector chains on original tuple terms
  - function leaves are rewritten with lambda wrappers so interpolants are expressed over original function terms and tuple arguments

### Backtracking/GC behavior
- Tuple-blast mappings and generated artifacts are integrated with preprocessor push/pop and GC marking.
- Original assertion structure remains separated from processed assertions; preprocessing data is scoped and cleaned with solver backtracking.

### Tests
- Added/expanded `tests/api/mcsat_tuples.c` with MCSAT-only guards (`#ifdef HAVE_MCSAT` + runtime `yices_has_mcsat()`).
- Coverage includes:
  - tuple-valued function model reconstruction
  - function update/application on tuple args/results
  - nested tuple + ite/distinct cases
  - push/pop scope behavior
  - UNSAT + interpolant retrieval on tuple/function constraints


src/mcsat/solver.c
```diff
@@ -3123,10 +3123,10 @@ void mcsat_stop_search(mcsat_solver_t* mcsat) {
 }
 
 term_t mcsat_get_unsat_model_interpolant(mcsat_solver_t* mcsat) {
-  return mcsat->interpolant;
+  return preprocessor_unblast_term(&mcsat->preprocessor, mcsat->interpolant);
```
ahmed-irfan: why do you need this? 
isn't mcat->interplant already unblasted?

disteph: Good catch. The code wasn't being very clean there. Fixed in commit [c7e51e8](https://github.com/SRI-CSL/yices2/pull/606/commits/c7e51e80c50612f291ac31e7d8c70f411597f4b1): Now mcsat->interpolant is in the blasted world. Asking for it via the getter does the unblasting,

ahmed-irfan (approved):

disteph:
Review by Windsurf/Opus4.7:

# Review: `7a3358d6` (`mcsat-tuples-support`) vs `master`

**Scope.** 15 commits, ~2,500 LOC added across `src/mcsat/{preprocessor,solver}.{c,h}`, `src/context/context_solver.c`, `src/mcsat/no_mcsat.c`, plus 1 large API test ([tests/api/mcsat_tuples.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:0:0-0:0), ~670 LOC) and 9 `.ys` gold regressions under `tests/regress/mcsat/tuples/`.

---

## High-level summary

The branch teaches the MCSAT preprocessor to handle tuple-typed terms (including tuples whose components are themselves functions), and threads the corresponding model-reconstruction and unsat-interpolant unblast through `mcsat_solver_t`. The new code is concentrated in one module ([preprocessor.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:0:0-0:0)), is well factored into a fixed pipeline (peek/get/set + iterative driver + memoized type-walks + unblast/build-model), and the public solver-side surface change is narrow: one renamed setter (`mcsat_set_unsat_result_from_labeled_interpolant`) and one new internal helper. Test coverage is substantial, with both API-level and gold `.ys`-level coverage including a deterministic `(show-model)` test.

---

## Strengths

### Architectural clarity

- **Single, documented invariant for the sticky interpolant.** `mcsat_solver_t::interpolant` is always in the public/post-unblast world. Two doc-comments — one on the field declaration in `solver.c`, one on the API on `solver.h` — describe the contract and force every writer through the right entry point. The `mcsat_set_interpolant_from_internal` helper centralizes the unblast call, and `false_term` literal writes are explicitly tagged `/* false_term is already in the public world */`, so a future reviewer can follow every interpolant store on one screen.

- **Tuple blasting is a well-bounded subsystem.** All new state is a contiguous block in `preprocessor_t` (the `tuple_blast_*` ivectors/maps and the three memoization caches), with matching `init_*` and `delete_*` calls and `scope_holder_push`/`pop` integration for backtracking. The packed `tuple_blast_data` layout (`[size, terms...]`) keeps lookup at O(1) and is what enables the zero-copy peek.

- **The iterative driver eliminates the stack-blowup risk.** Replacing a recursive [tuple_blast_term](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:1019:0-1091:1) with the explicit `preprocessing_stack`-based loop means an arbitrarily deep DAG cannot crash the solver via C stack. The [term_has_tuples_in_subdag](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:422:0-490:1) memo and the tuple-free fast path mean tuple-free sub-DAGs are blasted in O(1) per node, which in practice means the iterative driver is no slower than the previous recursive version on the common case.

### Correctness hygiene

- **Memoization with a stable-key argument.** The caches `type_is_tuple_free_cache`, `type_leaf_count_cache`, and `term_has_tuples_cache` are documented as never needing invalidation because type/term IDs are immutable for the lifetime of the table. That argument is correct in this codebase and saves the complexity of tying these caches to GC.

- **Error paths converge.** [merge_blasted_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2395:0-2533:1) was refactored to a single cleanup label (`done:`) with a `fail_reason` string set before each early exit and emitted at cleanup. This is a real readability win over the prior multi-`return null_value` form, and pairs with the trace diagnostics so a silent drop is no longer silent.

- **Trace diagnostics for silent drops.** [preprocessor_build_tuple_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2535:0-2613:1) now logs the dropped atom and the reason (no value in trail / no function value reconstructable) under the `mcsat::preprocess` trace tag. This addresses the residual-concern that an atom missing from `(show-model)` had no observable cause.

- **Strict UNSAT assertion preserved end-to-end.** The `efa22f5234` commit (after rebase) now contains both the legitimate test stabilization (`yices_default_config_for_logic(... "QF_UFLIA")`) and the strict `assert(... == YICES_STATUS_UNSAT)` form. The history no longer shows the weaken/restore oscillation.

### Performance

- **Zero-copy peek across all read-only single-leaf sites.** Every tuple-blasting case whose only use of the blasted snapshot is a `(size==1, data[0])` read uses [tuple_blast_peek](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:273:0-288:1) and bypasses the per-sub-term `malloc + memcpy + free` of the `_get` path: NOT, SELECT, EQ, ITE, DISTINCT, OR/XOR, POWER_PRODUCT, all four `*_POLY`, and the generic-composite combine. The peek doc-comment is explicit about the lifetime invariant and lists which call sites stay on the copying path (APP, UPDATE, top-level substitution, model reconstruction) and why.

### Tests

- **API test ([tests/api/mcsat_tuples.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:0:0-0:0)).** Covers tuple-of-int/bool, function-valued tuple components, `tuple-update` interactions, push/pop scope semantics, model reconstruction, and the no-residual-lambda invariant on interpolants. The no-lambda check directly addresses H2 — i.e., the worry that a function-tuple atom's interpolant could leak an internal lambda body into the user's term world.

- **`.ys` gold regressions.** Nine tests covering distinct/unsat, function-domain/range, nested ITEs, mixed selectors over arith/BV, update-rewrite, push/pop, and the new [tuple_show_model.ys](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/mcsat/tuples/tuple_show_model.ys:0:0-0:0) deterministic gold for `(show-model)` over plain `(int × bool)` tuples. The show-model test is verified byte-identical across debug, release, and sanitize.

### Build & test discipline

- All 31 API tests pass in **debug**, **release**, and **sanitize** (with the unrelated, pre-existing `type_macros` heap-UAF on master being the only sanitize failure, also reproducible on master HEAD).
- All 691 mcsat regression tests pass in all three modes.

---

## Concerns / things I would still want to discuss

### 1. [merge_blasted_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2395:0-2533:1): fail-on-decompose still silently degrades the model

The function returns `null_value` if any blasted leaf doesn't decompose cleanly to a function-or-update, the caller drops the atom, and the user sees the atom missing from `(show-model)`. The trace diagnostic is the right minimum, but a user running without a tracer enabled still gets a quietly under-specified model. **Suggestion:** consider promoting the dropped-atom set to a queryable property (e.g. a counter on the preprocessor), or at least `fprintf(stderr, ...)` when the drop happens with `--mcsat` in interactive mode. **Severity: medium.** Not a merge blocker.

### 2. The [tuple_blast_peek](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:273:0-288:1) lifetime invariant is fragile

It is stated clearly in the doc-comment, and every current caller respects it, but the rule ("MUST NOT hold the returned pointer across any operation that can grow `tuple_blast_data` -- most notably [tuple_blast_term](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:1019:0-1091:1)") is exactly the kind of invariant a future contributor will violate. **Suggestion:** add a debug-only generation counter on `tuple_blast_data` (incremented on every `ivector_add`/push that grows it), capture it in peek, and assert in a debug-only wrapper that the generation matches when the caller dereferences. Keeps the zero-copy benefit in release but makes a regression unmissable. **Severity: low.** Pure defensive engineering.

### 3. `mcsat_set_unsat_result_from_labeled_interpolant` precondition isn't enforced

The doc-comment states "interpolant is already in the public / post-unblast world" but no assertion checks it. If a future caller forgets and feeds an internal-world term, blasted leaves leak into the sticky interpolant. The unblast operation is idempotent, so a defensive `apply_term_subst` against the empty substitution wouldn't catch it, but a debug-only walk that asserts no internal-leaf variable appears would. **Severity: low.**

### 4. Test file is large (~670 LOC) and would benefit from a helper

[tests/api/mcsat_tuples.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:0:0-0:0) is a flat list of test functions, several of which share boilerplate (context setup, `make_mcsat_context(true/false)`, the no-residual-lambda check). The helpers exist (`assert_no_lambda_residue` and friends) but the per-test setup/teardown is still copy-pasted. **Severity: low**, cosmetic.

### 5. Sanitize fail (`type_macros`) is pre-existing on master, not from this branch

Verified by switching to `master` and reproducing. Not a branch-blocker, but worth filing a separate issue against `tests/api/type_macros.c` /`src/utils/symbol_tables.c`.

---

## Smaller observations

- The `// BD` comment removed from `mcsat_clear` is the right cleanup — it carried no information.
- The [term_has_tuples_in_subdag](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:422:0-490:1) cache is keyed on `index_of(t)` (polarity-insensitive), which is correct because tuple-ness doesn't depend on negation. The doc-comment says so. Good.
- [preprocessor_unblast_term](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2773:0-2825:1) is exposed in `preprocessor.h`. Since it's only consumed by `solver.c`, you could keep it `static` if you preferred a tighter surface, but the current factoring also makes it easier to test in isolation.
- `init_ivector(&eqs, n)` (and similar) in [tuple_blast_eq_vector](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:508:0-524:1) pre-sizes the ivector, which is the right idiom; consistent throughout the new code.
- The `.ys.options` files all carry just `--mcsat`. One could imagine a global default for `tests/regress/mcsat/tuples/`, but the current explicit per-file form matches the convention elsewhere.

---

## Verdict

**Approve, with the medium-severity suggestion on the silent-drop telemetry left as a follow-up.** The architecture is clean, the invariant for the public-world interpolant is documented and enforced through the type-of-call (writers must pick the right entry point), the iterative driver removes a real correctness risk, the memoization is well argued, and test coverage at both the C-API and `.ys` levels is thorough. The branch is in a state where I would be comfortable merging once you've decided how aggressively you want to expose the silent-drop signal to users.

src/mcsat/solver.c
disteph: Let me roll that back. It is cleaner to keep mcsat->interpolant in the pre-blast world where the user-written assertions and assumptions also belong. New commits enforce that.

disteph:
Code review by OpenCode/GPT5.5 about commit [995ec36](https://github.com/SRI-CSL/yices2/pull/606/commits/995ec36d497e5b2f94a80825add4a160ab647f6c):

Findings No current findings.

The prior issues are addressed:

H1 dispatch gap: fixed for missing term kinds.
H2 stale memo caches: reset on GC.
M1 lambda-residue risk: now covered by naked function-component interpolant regression.
M2 partial tuple model with keep_subst=0: source fix plus regression.
Ceil note: regression now actually exercises ARITH_CEIL dispatch with a lax constraint.
Verified:

mcsat_tuples passes in debug.
tests/regress/mcsat/tuples passes: 11/11.
git diff --check HEAD~3..HEAD clean.
Residual risk: the known MCSAT ceil semantic bug is real but out of scope for this PR.

disteph:
Summary of changes since Code review of 7a3358d6:

# Branch summary: `mcsat-tuples-support`, H2 onward

Five commits, currently `995ec36d..42bd5656^`. All target MCSAT's tuple-blasting infrastructure and its model-reconstruction edges.

## `42bd5656` — H2: reset tuple-blast memo caches on GC

The preprocessor maintains three opportunistic memoization caches keyed on raw term/type IDs (`type_is_tuple_free_cache`, `term_has_tuples_in_subdag_cache`, `term_kind_tuple_classification_cache`). The original code assumed those IDs were stable for the lifetime of the table. They are not — Yices recycles term and type IDs after a GC sweep (`indexed_table_free` under `src/terms`), so a freshly recycled ID can pick up a cached classification belonging to the freed term, which then drives the iterative tuple-blast dispatch down the wrong arm.

**Fix**: clear the three opportunistic memos at the start of [preprocessor_gc_mark](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:3082:0-3135:1). The `tuple_blast_map` / `preprocess_map` / `purification_map` are kept — their entries are explicitly marked alive, so they survive GC by construction.

Includes a regression test that forces a GC between two preprocessing rounds with overlapping ID space.

## `99293ae0` — H1: tuple-blast dispatch for missing term kinds

[term_has_tuples_in_subdag](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:453:0-521:1) did not descend through several unary/binary arithmetic kinds (`ARITH_IS_INT_ATOM`, `ARITH_FLOOR`, `ARITH_CEIL`, `ARITH_ABS`, `ARITH_FF_EQ_ATOM`, `ARITH_FF_BINEQ_ATOM`) and the bit-extraction kind `BIT_TERM`. Result: `(<op> (select x i))` was misclassified as tuple-free, and an un-blasted `SELECT` reached the legacy preprocessor, producing a `formula not supported` error.

**Fix**: add dispatch entries in both the descent helper and the rebuild loop, so the operator is rebuilt over the blasted operand.

Coverage:
- `BIT_TERM` over a tuple-component bv: exercised at the C-API level in [tests/api/mcsat_tuples.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:0:0-0:0) ([test_tuple_blast_bit_over_tuple_component](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:787:0-816:1)).
- `ARITH_FLOOR` / `ARITH_IS_INT_ATOM` / `ARITH_CEIL` / `ARITH_ABS`: exercised at the `.ys` level in [tests/regress/mcsat/tuples/tuple_unary_arith.ys](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/mcsat/tuples/tuple_unary_arith.ys:0:0-0:0).
- `ARITH_FF_EQ_ATOM` / `ARITH_FF_BINEQ_ATOM`: cannot be exercised end-to-end (FF arithmetic isn't reachable from the `.ys` or SMT2 frontends combined with tuples). Reviewed by inspection against the binary-arith dispatch shape, which is structurally identical.

## `51121a26` — M1: regression test for naked function-valued tuple component

Pure guard-rail test, no source change. Pins the property that interpolants over function-valued tuple components remain lambda-free, so future changes that start producing naked `(lambda (v0) ...)` residues on the `(b)` path (multi-leaf function ranges where the per-slice lambda body is non-eta-reducible) will fail this test rather than user-facing call sites.

The "naked" path is currently doubly guarded:
- Single-leaf function ranges: [mk_lambda](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/term_manager.c:4431:0-4450:1) eta-reduces `(lambda (v0) (f v0))` → `f` at construction.
- Multi-leaf function ranges: MCSAT interpolation in this configuration returns the original asserted formula rather than a leaf-level rewrite, so unblast never substitutes the leaf inside the interpolant.

Test asserts `(= fx fy)` against `(distinct fx fy)` over a function whose range is itself a tuple, then runs [assert_interpolant_has_no_lambda_residue](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:529:0-537:1) on the result.

## `6c13fcf2` — M2: preserve tuple atoms under `keep_subst=0` model builds

Two related fixes in [preprocessor_build_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:3000:0-3053:1) and [preprocessor_build_tuple_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2674:0-2765:1), both triggered when `yices_get_model(ctx, 0)` builds a public model with `has_alias=false`.

1. [preprocessor_build_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:3000:0-3053:1) previously called `model_add_substitution` unconditionally for every solved equality. With `keep_subst=0` that asserts `has_alias` in debug builds and silently drops the binding in release. **Fix**: split on `model->has_alias`. When false, lazily initialize an evaluator, evaluate `eq_subst` to a concrete value, and `model_map_term(eq_var, v)`. Falls back to `vtbl_make_object` if `eval_in_model` returns an error (e.g., `eq_subst` transitively depends on another not-yet-mapped `eq_var`).

2. [preprocessor_build_tuple_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2674:0-2765:1) previously dropped a tuple atom whenever any blasted leaf had no trail value — typical for unconstrained tuple components, made fatal by `keep_subst=0` disabling the alias-based default-completion in `eval_uninterpreted`. **Fix**: fall back to `vtbl_make_object` on the leaf's declared type so the parent tuple atom is reconstructed instead of dropped. Truly uninhabitable types (shouldn't occur in well-formed input) still take the drop path, with a more accurate trace message.

Includes [test_partial_tuple_model_no_keep_subst](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:720:0-782:1), which asserts `(= (select x 1) 5)` over `x : (tuple int int)`, builds the model with `keep_subst=0`, and checks both that component 1 = 5 and that the whole tuple reconstructs.

## `995ec36d` — actually exercise ARITH_CEIL dispatch in [tuple_unary_arith.ys](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/mcsat/tuples/tuple_unary_arith.ys:0:0-0:0)

Doc-and-test cleanup, no source change. The original [tuple_unary_arith.ys](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/mcsat/tuples/tuple_unary_arith.ys:0:0-0:0) header claimed it covered ARITH_CEIL but no `ceil` assertion was present, on the (incorrect) assumption that "mcsat's nra plugin cannot solve any `(ceil x) = k`."

In fact mcsat does route and answer ceil queries — but answers are wrong: for non-integer rationals it computes `floor` instead of `ceil`. That's a separate, pre-existing semantic bug in [src/mcsat/preprocessor.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:0:0-0:0) (the `ARITH_CEIL` arm builds `arith_floor` after purification — a 2016 copy-paste from the FLOOR arm). It is out of scope for this PR; will be filed as a separate issue.

To still exercise ARITH_CEIL dispatch routing here, the test asserts a deliberately *lax* `(<= (ceil (select x 2)) 100)` constraint that holds under both the correct and the buggy interpretation. The matching note in [tests/api/mcsat_tuples.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:0:0-0:0) was updated to point at the `.ys` header for the ceil-semantics caveat.

## Pipeline status

- Local: `make MODE=debug check-api` 31/31 pass; `make MODE=release check-api` 31/31 pass. All 11 `tests/regress/mcsat/tuples/*.ys` regressions pass. Full `make MODE=debug check` has 2 unrelated pre-existing failures (`set-timeout.smt2` flaky timing, `wd/issue401b.smt2` 60s timeout) that reproduce on the previous `7a3358d6` tip without these commits.
- CI: `-Werror=comment` build-break in the original H2 was fixed in place via `git rebase --autosquash` and force-pushed; H2 now builds clean with strict comment checking.

disteph:
### Review of `b4507500` by Windsurf/Opus4.7 — *mcsat: avoid evaluating unmapped tuple-blast leaves*

**What it does.** Seven added lines in [preprocessor_build_tuple_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:2701:0-2817:1). Per blasted leaf, try a direct `model_find_term_value` first, and only fall through to [model_get_term_value](cci:1://file:///Users/sgl/git/yices/yices2/src/model/model_queries.c:28:0-52:1) (which can invoke [eval_in_model](cci:1://file:///Users/sgl/git/yices/yices2/src/model/model_eval.c:1910:0-1931:1)) when the evaluator has a chance of producing a value: either an alias table is present, or the leaf isn't an unmapped `UNINTERPRETED_TERM`. Otherwise we keep `v == null_value` and let the existing `vtbl_make_object` fallback mint a well-typed default.

**Root cause it fixes.** [test_partial_tuple_model_no_keep_subst](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/mcsat_tuples.c:733:0-820:1) builds a tuple-typed uninterpreted `x`, constrains only the first component, and calls `yices_get_model(ctx, /*keep_subst=*/0)`. The unconstrained second leaf is never assigned during search; the resulting model has `has_alias == false`. The old code therefore entered [eval_uninterpreted](cci:1://file:///Users/sgl/git/yices/yices2/src/model/model_eval.c:1601:0-1622:1) and took the `MDL_EVAL_UNKNOWN_TERM` exit via `longjmp` to the `setjmp` in [eval_in_model](cci:1://file:///Users/sgl/git/yices/yices2/src/model/model_eval.c:1910:0-1931:1). That control flow is correct and longstanding (the evaluator has used `setjmp`/`longjmp` for unevaluable terms since 2010/2013; nothing new in this PR), but on `windows-latest | release | --enable-thread-safety --enable-mcsat` the unwind itself crashes — empirically requiring all three of `-O3`, `-fomit-frame-pointer`, and `-DTHREAD_SAFE`. That's a known-fragile mingw-w64 SEH-based `setjmp`/`longjmp` interaction, not a Yices bug. The patch sidesteps it on the only path the API tests reach it.

**Why this shape.** Semantics-preserving (the `v < 0` branch already minted defaults for this case), minimal (+7/-2 in one file), localized to the new tuple-blast model build, and a net perf wash on the common path (the extra lookup replaces the one [model_get_term_value](cci:1://file:///Users/sgl/git/yices/yices2/src/model/model_queries.c:28:0-52:1) would have done internally before falling through to the evaluator).

**Suggested follow-up (separate issue).** Other Yices entry points that reach [eval_in_model](cci:1://file:///Users/sgl/git/yices/yices2/src/model/model_eval.c:1910:0-1931:1) on unevaluable terms (free vars, lambdas, quantifiers, algebraic atoms) remain vulnerable on the same MinGW toolchain. The robust fixes are (a) build Windows release with `-fno-omit-frame-pointer`, or (b) replace the `setjmp`/`longjmp` channel in [model_eval.c](cci:7://file:///Users/sgl/git/yices/yices2/src/model/model_eval.c:0:0-0:0) with explicit error returns. Out of scope here.

LGTM.

disteph:
Code review for entire PR by Codex/GPT5.5:

**Findings**
No correctness, API compatibility, or merge-blocking issues found in `HEAD` (`9f29aab3`) compared with `master` (`7878664`).

**Summary**
The PR adds MCSAT tuple support by tuple-blasting public tuple/function terms into scalar leaves inside `src/mcsat/preprocessor.c`, then reconstructing public-facing models and interpolants before exposing them through the API. The solver-side interpolant contract is explicit: internal conflict-analysis interpolants are unblasted before storage, while assumption-label substitution is handled by `mcsat_set_unsat_result_from_labeled_interpolant`.

The test coverage is substantial for the changed behavior: API tests cover tuple/function models, tuple selectors under arithmetic/BV contexts, push/pop scoping, GC, `keep_subst=0` model reconstruction, and interpolation cases including function-valued tuple components. The regression tests add frontend coverage for tuple domains/ranges, distinct, update, nested ITE, selectors, bit selection, model display, and unary arithmetic dispatch.

**Tests Run**
- `make -s check-api`

Result: 31 passed, 0 failed.

PR #614, opened 2026-03-11
Fix #613 MCSAT curried function model reconstruction

### LINKED ISSUE #613: MCSAT crashes when generating models for curried functions
Hello,

We notices some crashes in `mcsat` when using curried functions:
```c++
#include <cassert>
#include <iostream>

#include <yices.h>

context_t *newContext(const char *solver) {
    auto cfg = yices_new_config();
    yices_set_config(cfg, "solver-type", solver);
    yices_set_config(cfg, "mode", "interactive");
    yices_set_config(cfg, "model-interpolation", "true");
    auto context = yices_new_context(cfg);
    yices_free_config(cfg);
    return context;
}

int main() {
    yices_init();
    auto integer = yices_int_type();

    auto array = yices_new_uninterpreted_term(yices_function_type1(integer, yices_function_type1(integer, integer)));
    yices_set_term_name(array, "array");

    // array[1][7] = 10
    auto f = yices_eq(
        yices_application(yices_application(array, 1, (int[]){yices_int32(1)}), 1, (int[]){yices_int32(7)}),
        yices_int32(10));

    auto ctx = newContext("mcsat");
    yices_assert_formula(ctx, f);

    auto r = yices_check_context(ctx, nullptr);
    assert(r == YICES_STATUS_SAT);

    std::cout << "SAT" << std::endl;
    auto model = yices_get_model(ctx, true);
    std::cout << yices_model_to_string(model, 100, 80, 0) << std::endl;

    yices_free_context(ctx);
    yices_exit();
    return 0;
}
```
There are no crashes when using `dpllt` as solver, and a correct model is printed:
```
(function array
 (type (-> int (-> int int)))
 (= (array 1) fun!6)
 (default fun!9))
(function fun!6
 (type (-> int int))
 (= (fun!6 7) 10)
 (default 0))
(function fun!9
 (type (-> int int))
 (default 2))
```
We're using functions in Yices to handle Smtlib [Arrays](https://smt-lib.org/theories-ArraysEx.shtml). Most of the time only one index is needed, and that case works just fine with MCSAT. However, if MCSAT fails we'd prefer to get a proper error result, instead of a segfault

disteph:
Fixes #613.

This fixes an MCSAT model-construction bug on curried functions such as `array(1)(7)`, which could reach `sat` and then crash while building/printing the model. The UF model builder now reconstructs nested function values recursively from the MCSAT trail instead of treating function-valued intermediates as first-order defaults.

This PR also adds a focused API regression test for the reported curried-function case, fixes an MCSAT-specific lambda error-path mismatch in the SMT2 preprocessor, and extends the regression harness with a `tests/regress/both` bucket that runs selected higher-order/array regressions under both `--mcsat` and `--dpllt`.


disteph:
Codex/gpt5.5 code review:

No findings.

**Open Questions / Assumptions**

I reviewed the committed branch `issue-613-mcsat-curried-function-model` against `master`. The two untracked files under `tests/api/` are unrelated local work and are not part of this PR.

**Review Summary**

The PR looks ready to merge.

The core change in [uf_plugin.c](/Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c) fixes the real model-construction bug: MCSAT now reconstructs function values from trail-assigned function ids and recursively handles function-valued results, which is the right shape for curried terms like `array(1)(7)`.

The related printer change in [smt2_printer.c](/Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c) is also justified. Copying update-map ids out of `table->hset1` before recursive printing removes a concrete reentrancy hazard without changing model semantics.

The test coverage is now adequate for the merge:
- [mcsat_curried_function_model.c](/Users/sgl/git/yices/yices2/tests/api/mcsat_curried_function_model.c) covers the original API reproducer and checks the nested value semantically.
- [issue613_model_print.smt2](/Users/sgl/git/yices/yices2/tests/regress/mcsat/arrays/issue613_model_print.smt2) directly exercises the SMT2 `get-model` / `get-value` printing path for nested arrays.
- [run_test.sh](/Users/sgl/git/yices/yices2/tests/regress/run_test.sh) and [README.md](/Users/sgl/git/yices/yices2/tests/regress/both/README.md) now describe the shared `both` behavior accurately: MCSAT plus the default DPLL(T) path.

Residual risk is limited to the normal breadth of MCSAT UF model reconstruction: this changes all function-typed UF model building in that plugin, not only the exact reproducer. I did not see an obvious correctness issue in the implementation or a missing targeted regression for the reported failure mode.

**Tests Run**

- `build/arm-apple-darwin23.6.0-debug/bin/mcsat_curried_function_model`
- `tests/regress/check.sh tests/regress/both build/arm-apple-darwin23.6.0-debug/bin`
- `tests/regress/run_test.sh tests/regress/mcsat/arrays/issue613_model_print.smt2 build/arm-apple-darwin23.6.0-debug/bin`
- `build/arm-apple-darwin23.6.0-debug/bin/yices_smt2 --mcsat tests/regress/coverage/random/QF_AUFLIA/2328a12f.smt2`
- `build/arm-apple-darwin23.6.0-debug/bin/yices_smt2 --mcsat tests/regress/wd/issue404.smt2`

disteph:
Code review by Codex/GPT5.5 on the additional 2 commits:

No findings.

**Open Questions / Assumptions**

I reviewed exactly the last two commits on `issue-613-mcsat-curried-function-model`:

- `802c426a` `Fix use-after-free and related issues in MCSAT curried function model`
- `8e314f4a` `Add stress regression for MCSAT curried function model int_hmap resize`

The worktree is clean.

**Review Summary**

The hash-map fix in [uf_plugin.c](/Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:831) looks correct. The old risk was holding an `int_hmap_pair_t *` across recursive calls that may insert into `builder->function_value` and resize the map. The new code caches the base function before recursion, then re-fetches the entry before storing the final updated function value at [uf_plugin.c](/Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:878). I did not find another equivalent stale-pointer path in the changed builder code.

The concrete value printer change in [concrete_value_printer.c](/Users/sgl/git/yices/yices2/src/io/concrete_value_printer.c:44) matches the earlier SMT2 printer fix: copy `table->hset1` contents before recursive printing can clobber that scratch set. The cleanup is correct; `safe_free(NULL)` is valid in this codebase.

The new stress regression [issue613_many_curried.smt2](/Users/sgl/git/yices/yices2/tests/regress/mcsat/arrays/issue613_many_curried.smt2) is well targeted. It exercises enough curried array terms to force `int_hmap` resizing during model reconstruction, which is the failure mode fixed by the first commit.

One note: [mcsat_curried_function_model.c](/Users/sgl/git/yices/yices2/tests/api/mcsat_curried_function_model.c) now relies entirely on `assert` for the semantic checks again, but this file explicitly undefines `NDEBUG`, consistent with several existing API tests, so I don’t see that as a defect.

**Tests Run**

- `git diff HEAD~2..HEAD --check`
- `make MODE=debug check-api`
- `tests/regress/run_test.sh tests/regress/mcsat/arrays/issue613_many_curried.smt2 build/arm-apple-darwin23.6.0-debug/bin`
- `tests/regress/run_test.sh tests/regress/mcsat/arrays/issue613_model_print.smt2 build/arm-apple-darwin23.6.0-debug/bin`
- `tests/regress/check.sh tests/regress/both build/arm-apple-darwin23.6.0-debug/bin`

disteph:
Code review by Windsurf/Opus4.7 high:

# Code Review: `issue-613-mcsat-curried-function-model` → `master`

**Base:** `895707f2` (master) · **Head:** `8e314f4a` · **Δ:** +708 / −213 across 34 files · **Commits:** 8

## Scope and intent

The PR addresses GitHub issue #613: MCSAT produces wrong or crashing models for terms whose head is itself a function application (curried use of higher-order arrays, e.g. `(select (select a i) j)` where `a : (Array Int (Array Int Int))`). Before this change, [uf_plugin_build_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:945:0-1066:1) assumed every application had an uninterpreted head symbol. The fix replaces that assumption with a function-*identity* (value-based) grouping and builds per-function values recursively via `vtbl_mk_update` over a fresh base.

The PR additionally:

- Hardens two concrete-value printers against a `hset1` reentrancy hazard that is only exposed when printing nested functions.
- Raises a clearer exception on `LAMBDA_TERM` in the MCSAT preprocessor.
- Introduces a `tests/regress/both/` category that exercises each test under both solver back-ends.
- Adds SMT2 regressions for issue #613 (one basic, one stress) and an API-level regression.

## Commit-by-commit summary

| Hash | Author | Summary |
|---|---|---|
| `b2c49c80` | original | Core MCSAT fix + printer hardening + preprocessor exception + API test scaffolding. |
| `97945b58` | original | `run_test.sh` / `check.sh` changes to run `tests/regress/both/` in both modes. |
| `6f3dc09a` | original | Keep the new API test working when `NDEBUG` strips `assert`. |
| `dc5687ca` | original | Reinstate assertions in the API test. |
| `4f83bf00` | original | Merge `master` into the branch; no functional changes. |
| `d6878785` | original | `issue613_model_print.smt2` regression (single curried array). |
| `802c426a` | this review | Fix use-after-free in [uf_model_builder_get_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:816:0-896:1), finite-domain fallback, div/mod dead-code cleanup, printer reentrancy finalization. |
| `8e314f4a` | this review | Stress regression (`issue613_many_curried.smt2`) that deterministically reproduces the use-after-free under ASan. |

## Detailed review

### [src/mcsat/uf/uf_plugin.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:0:0-0:0) (+441 / −110)

The heart of the PR. A new `uf_model_builder_t` owns all per-build state (fresh-value maker, `app_terms` ivector, `arguments` buffer, plus three `int_hmap`s for `app_term_first`, `function_term`, and `function_value`). [uf_plugin_build_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:945:0-1066:1) now executes in two clearly separated phases:

1. **Uninterpreted-function phase.** Iterate the trail; for every top-level variable whose term is an `UNINTERPRETED_TERM` of function type, call [uf_model_builder_get_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:816:0-896:1) and map it in the model.
2. **Special phase.** Filter the eq-graph trail for `ARITH_RDIV`/`ARITH_IDIV`/`ARITH_MOD`, sort by kind, and emit one function value per kind for the division-by-zero mappings.

The grouping by **function identity** rather than by **head term** is the correct fix for #613. The identity is the `int32_t` assignment the trail holds for the function variable ([uf_plugin_get_function_id_from_trail](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:662:0-677:1)), which is shared by any two syntactically distinct terms that denote the same function at model-build time. This is precisely what the curried case needs: `(select a i)` and `(select a j)` may or may not share identity depending on the trail.

#### Strengths

- Clear separation of concerns: `_construct`/`_destruct`/`_prepare`/`_get_function_value`/`_get_term_value`/`_remember_function_term` each do one thing.
- [uf_model_builder_prepare](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:897:0-933:1) precomputes the `app_term_first` index, so the inner loop in `_get_function_value` is O(#apps-for-this-f) rather than O(total apps).
- The recursive structure handles arbitrary nesting correctly: a function-typed argument is resolved by `_get_term_value` → `_get_function_value`, which cycles back through the cache keyed on function-id.

#### Issues fixed during review (commit `802c426a`)

1. **Heap-use-after-free across `int_hmap` resize.** The pre-review code cached a pointer into `builder->function_value` before recursing and wrote through it on return:

   ```c
   find = int_hmap_get(&builder->function_value, function_id);
   ...
   for (...) { ... uf_model_builder_get_term_value(...); ... }   // may call int_hmap_get internally
   find->val = f_value;                                           // UAF if int_hmap_extend ran
   ```

   `int_hmap_get` calls `int_hmap_extend` once the live-count crosses the load-factor threshold, which `safe_free`s the old bucket array. AddressSanitizer confirms the use-after-free on the stress test:

   ```
   WRITE of size 4 at 0x61100001fca4 (uf_model_builder_get_function_value)
   freed by thread T0:
     int_hmap_extend int_hash_map.c:173
     int_hmap_get   int_hash_map.c:265
     uf_model_builder_get_function_value (recursive)
   ```

   Fixed by caching the value early, setting `find = NULL` to document invalidation, and re-fetching via `int_hmap_get` after the recursive loop. The re-fetch is cheap (one hash probe) and the only alternative — tracking "will any recursion trigger `extend`" — is brittle.

2. **`assert(f_value != null_value)` on finite saturation.** [make_fresh_function](cci:1://file:///Users/sgl/git/yices/yices2/src/model/fresh_value_maker.c:610:0-714:1) legitimately returns `null_value` for finite-domain/finite-range pairs once every function of that type has been materialized. This asserts in debug and is UB in release. Replaced with a fall-through that constructs a constant function with a default range value — keeping model construction total. This is the same defensive strategy the printer-side code already uses ([vtbl_mk_default](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:750:0-755:1)).

3. **Dead branches in the div/mod pass.** [uf_plugin_build_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:945:0-1066:1)'s tail loop carried `app_f`, `prev_app_f`, `app_construct`, and an `APP_TERM` case that can never fire (the special loop only inserts `ARITH_*` kinds). Simplified to a `has_prev` flag plus the three real kinds; an `assert` at the loop head documents the invariant. No behavior change.

4. **[vtbl_mk_default(type_table_t*, value_table_t*, type_t)](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:750:0-755:1) → [vtbl_mk_default(value_table_t*, type_t)](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:750:0-755:1).** The first parameter was unused (the body just called `vtbl_make_object`, which already looks up `vtbl->type_table`). Signature simplified and all call sites updated.

#### Residual comments / minor

- [uf_model_builder_remember_function_term](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:790:0-811:1) documents a "keep the first `UNINTERPRETED_TERM` we see" priority rule (commit `802c426a` adds the explanatory comment). Good, but the logic deserves a brief note in the header about *why* (model printing prefers the declared symbol). Optional.
- [uf_model_builder_get_term_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:898:0-905:1) is a forward declaration (`static value_t uf_model_builder_get_term_value(...)`) followed by the mutually-recursive pair. Worth keeping the forward decl adjacent to the definition — currently they are 100 lines apart, which is fine but a `// defined below` comment at the declaration would help.
- The `ivector_t arguments` inside [uf_model_builder_get_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:816:0-896:1) is allocated and freed on every call. `builder->arguments` already exists for exactly this role and is used elsewhere. Could be reused to avoid churn on deeply-nested curried chains. Not a correctness issue; nice-to-have.

### `src/frontend/smt2/smt2_printer.c` and `src/io/concrete_value_printer.c` (+131 / −46)

Three print-update sites (`smt2_normalize_and_pp_update`, `smt2_pp_array_update`, `smt2_pp_update_definition` in the SMT2 printer; `vtbl_normalize_and_print_update`, `vtbl_normalize_and_pp_update` in the native printer) all did:

```c
vtbl_expand_update(table, c, &def, &tau);   // writes into table->hset1
hset = table->hset1;
for (i = 0; i < hset->nelems; i++) {
    mp = vtbl_map(table, hset->data[i]);
    smt2_pp_object(...);                    // may recurse → vtbl_expand_update → clobbers hset1
}
```

With curried ranges, the inner `smt2_pp_object` prints a nested update and overwrites `hset1` mid-iteration. The outer loop then reads stale indices.

The fix (`copy_update_maps` helper, introduced identically in both files) expands once, deep-copies `hset->data[0..n-1]` into a `safe_malloc`'d array, and iterates that. The print recursion can now clobber `hset1` freely; the caller is isolated.

#### Strengths

- Minimal, local, obviously correct.
- Added in both printers with the same signature — easy to keep in sync.

#### Issues / nits

- **Code duplication.** `copy_update_maps` is duplicated verbatim between the two printer files. Given both include `"model/concrete_values.h"`, a better home is `concrete_values.c` as `vtbl_copy_update_maps(...)` (or a static file-scope helper migrated to a shared header). Would reduce two copies to zero without changing semantics. Recommend for a follow-up.
- `copy_update_maps` returns `NULL` and sets `*n = 0` for empty maps; every caller handles that correctly via the `for (i=0; i<n; i++)` bound, but the `safe_free(NULL)` at function end relies on glibc/musl/`safe_free` tolerating NULL (it does in this codebase — `src/utils/memalloc.c` `safe_free` checks for NULL). Worth one comment at the helper.

### `src/mcsat/preprocessor.c` (+4 / −1)

```c
case LAMBDA_TERM:
    longjmp(*pre->exception, LAMBDAS_NOT_SUPPORTED);
    break;
```

Previously these fell through to `MCSAT_EXCEPTION_UNSUPPORTED_THEORY`, which obscures the diagnosis. Trivially correct. `LAMBDAS_NOT_SUPPORTED` is an already-defined exception code, so no follow-up work is needed downstream.

### `tests/regress/both/` and harness changes

- `tests/regress/both/README.md` clearly documents the dual-mode contract.
- The harness changes in `tests/regress/run_test.sh` and `tests/regress/check.sh` run each `.smt2` in `both/` twice, once with `--mcsat` and once with the default. The file naming of the moved tests (e.g. `wd_issue238.smt2`) is consistent.
- Two new smoke tests (`smoke_sat.smt2`, `smoke_unsat.smt2`) are useful sanity checks for the dual-mode runner itself.

One thing worth confirming before merge: that the CI matrix is actually set up to exercise this directory (i.e. the harness's new branches are hit in CI, not only locally).

### `tests/api/mcsat_curried_function_model.c`

End-to-end API coverage via the Yices C API: creates an MCSAT context, asserts a curried equality, retrieves the model, and walks the function values. Commits `6f3dc09a` and `dc5687ca` ping-ponged on NDEBUG safety. The current state (asserts only, no fall-through early-returns) is the right balance for a test binary — it will fail loudly under either build mode. The review-time cleanup (commit `802c426a` touches only comment and assert structure) doesn't change behavior.

### Regression tests

- **`tests/regress/mcsat/arrays/issue613_model_print.smt2`** — Single-array smoke test that covers `get-model`, `get-value` on the outer array, the curried intermediate, and the innermost scalar. Fails on master (heap-buffer-overflow in `vtbl_mk_function`), passes on the fix. Gold file is tight.

- **`tests/regress/mcsat/arrays/issue613_many_curried.smt2`** — 25 curried arrays. Two function ids per array (outer and the selected inner) put the `function_value` `int_hmap` across the 0.6×32 ≈ 19 growth threshold twice during model construction, which is what forces the `int_hmap_extend` realloc during the recursion. The test is precisely sized to the internal table parameters and has a short, deterministic gold. This is the only test in the suite that specifically guards the use-after-free: pre-fix it crashes under ASan on the write at [uf_model_builder_get_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:816:0-896:1) line 873 with a clean "freed by thread T0 … int_hmap_extend" trace.

  If `INT_HMAP_DEFAULT_SIZE` or the load factor ever change, this test's sensitivity will drift. A comment to that effect is in the `.smt2` header, but a stronger guardrail would be a compile-time `_Static_assert` somewhere binding the test's scale to those constants. Nice-to-have, not required.

## Correctness and safety

| Concern | Status |
|---|---|
| Original issue #613 (wrong/crashing model for curried heads) | Fixed; covered by two regressions. |
| `int_hmap_pair_t*` used across `int_hmap_get` / `int_hmap_extend` | Fixed in [uf_model_builder_get_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:816:0-896:1); ASan-verified. |
| `assert(f_value != null_value)` on finite-type saturation | Fallback path added. |
| `table->hset1` reentrancy in value printers | Eliminated via deep-copy helper. |
| Unsupported lambdas surfacing generically | Now raises `LAMBDAS_NOT_SUPPORTED`. |
| MCSAT + dual-mode regression coverage | `tests/regress/both/` plus harness support. |

## Test results

- `make check-api` → 28 / 28 pass.
- `make mcsat-check` → 686 / 686 pass (including both new SMT2 regressions).
- `REGRESS_FILTER=/both/ make check` → 9 / 9 pass.
- `MODE=sanitize` on the two new regressions and the trivial MCSAT smokes → all clean (no ASan / UBSan reports reachable from model construction).

## Suggested follow-ups (non-blocking)

1. Deduplicate `copy_update_maps` into a `vtbl_copy_update_maps` in `concrete_values.[ch]`.
2. Reuse `builder->arguments` instead of the per-call `ivector_t arguments` in [uf_model_builder_get_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:816:0-896:1) to reduce allocator churn on deep curry chains.
3. Add a brief architectural note above [uf_plugin_build_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/uf/uf_plugin.c:945:0-1066:1) explaining the two-phase (function / div-mod) split, mirroring the inline comments that `802c426a` added.
4. Consider binding `issue613_many_curried.smt2`'s array count to `INT_HMAP_DEFAULT_SIZE` via a compile-time constant so future `int_hmap` retunings don't silently lose the regression's ability to trigger a resize.
5. Confirm CI runs `tests/regress/both/` (not just locally).

## Recommendation

**Approve / merge.** The core fix is correct and well-structured, the secondary fixes (printer reentrancy, lambda exception, test-harness dual-mode) are independently valuable, the new regressions are both ASan-verifiable and tightly scoped, and the follow-ups are all cosmetic or future-proofing rather than correctness-blocking.

PR #623, opened 2026-05-05
smt2: print algebraic numbers in SMT-COMP 2026 root-of-with-interval …

ahmed-irfan:
…format

Previously, algebraic model values were printed as lossy float approximations (e.g. 1.414214). The SMT-COMP 2026 model validation track requires the root-of-with-interval format, which encodes the exact value as a polynomial plus an isolating interval.

The new smt2_pp_algebraic normalizes the polynomial at print time (square-free part via gcd(f,f'), then primitive part for coprime coefficients with positive leading coefficient) and formats it as

  (root-of-with-interval (coeffs p_0 ... p_n) min max)

where coefficients are listed in ascending degree order and bounds are SMT2 real literals (a.0 or (/ a.0 2^k.0) for dyadic rationals). Rational algebraic values (f == NULL in libpoly) are printed directly as real literals. Normalization happens only when a model is printed, so solving performance is unaffected.

Two regression tests are added: one for a proper algebraic value (√2) and one for the rational branch (x²=4, x>0 → 2.0). Four existing gold files are updated to the new format.


disteph (commented):
Review provided by Windsurf/Opus4.7

# Code review — PR #623 (smt2-algebraic-model-format)

**Baseline:** `master` ↔ `40fdd027` "smt2: print algebraic numbers in SMT-COMP 2026 root-of-with-interval format"
**Scope:** 1 source file (`src/frontend/smt2/smt2_printer.c:25-263`), 2 new regression tests, 4 updated `.gold` files.

## Overall

The change is small, well-isolated to the SMT2 printer, and mathematically sound. The square-free + primitive-part normalization is the right move (canonicalizes output, guarantees positive leading coefficient, no impact on solving). MCSAT-only (`#ifdef HAVE_MCSAT`) and print-time only — solver hot paths are untouched. **I'd merge it after addressing a few small points below.**

## Correctness

- **Square-free reduction** via `f / gcd(f, f')` and **primitive part** are correct. Per `poly/upolynomial.h:291-306`, `lp_upolynomial_primitive_part_Z` divides by `gcd(coeffs)` *with the sign of the leading coefficient*, so the output coefficients are coprime and the leading one is positive. The doc comment in the PR matches this. Good.
- **Rational branch** (`an->f == NULL`) matches libpoly's documented invariant in `poly/algebraic_number.h:30-40` ("If f is 0, then the interval is a single point"). So printing `I.a` is correct.
- **`pp_clone_string` is the right primitive** here: `src/io/yices_pp.c:901-906` confirms it makes an internal copy, so freeing `buf` immediately afterward is safe. ✅
- **Allocation accounting** in the proper-algebraic branch is balanced: `f_deriv`, `g`, optionally `f_sqfree`, `f_prim` all freed; each `coeffs[i]` is `lp_integer_construct`'d and `_destruct`'d. No leak.

## Issues / concerns

**1. Open vs. closed interval bounds are silently dropped.**
`lp_dyadic_interval_t` carries `a_open` / `b_open` bits (`poly/dyadic_interval.h:37-45`). The new code prints `I.a` and `I.b` without ever inspecting those bits. For libpoly's *isolating* intervals this is conventionally `(a, b)` open, but the SMT-COMP 2026 spec presumably defines a fixed convention. Worth either:

- adding a one-line comment in `smt2_pp_algebraic` stating "libpoly always returns an open isolating interval, which matches root-of-with-interval semantics"; or
- asserting `an->I.a_open && an->I.b_open` so we'd notice if libpoly ever changes that.

Either is fine; the silent drop is the only thing to fix.

**2. `q->n` (`size_t`) → `unsigned long` truncation, plus `mpz_ui_pow_ui` argument type.**

```c
  unsigned long k = q->n;
  ...
    mpz_ui_pow_ui(pow2, 2, k);
```

On LLP64 (64-bit Windows MinGW, which [windows_ci.yml](cci:7://file:///Users/sgl/git/yices/yices2/.github/workflows/windows_ci.yml:0:0-0:0) builds) `unsigned long` is 32 bits and `size_t` is 64. Truncation is a non-issue in practice (dyadic exponents from libpoly are never `> 2^32`), but for portability hygiene either cast `(unsigned long)q->n` explicitly or use `mp_bitcnt_t`. One-line nit.

**3. `free()` on memory returned by libpoly / GMP.**
Both `lp_integer_to_string()` and `mpz_get_str(NULL, ...)` allocate via the GMP allocator. `free()` matches only when the GMP allocator hasn't been overridden via `mp_set_memory_functions`. Yices doesn't override it, and other call sites (e.g. `string_buffer_append_mpz`) use the same idiom, so this is consistent — but worth a one-line comment "matches default GMP allocator" for the next reader, or move to `mp_get_memory_functions`-based release.

**4. `root-of-with-interval` is not stock SMT-LIB.**
It is the SMT-COMP 2026 model-validation track convention. Right now there's no in-tree comment pointing readers at the spec. A one-liner above `smt2_pp_algebraic` citing the SMT-COMP page (or the issue/PR that introduced the format) would save future archaeology.

**5. Backward-incompatible output.**
This silently changes what `(get-value ...)` and `(get-model)` print whenever an algebraic value escapes — any downstream consumer parsing Yices SMT2 output (CI of dependent projects, scripts, prior SMT-COMP runners) will see a different shape. Worth:

- a `RELEASE_NOTES` / `NEWS` entry; and
- ideally an opt-out (e.g. a `--algebraic-format=decimal|root-of` flag, or driven by `(set-option :produce-models …)` style). A flag is *not* required for merge IMO, but at least flag this in the PR description.

**6. No pretty-printer line-wrapping.**
Because the whole `(root-of-with-interval …)` is pushed as a single `pp_clone_string` token, the pretty printer can't break inside it. `tests/regress/mcsat/nra/model-eval.smt2.gold:8-9` already shows ~165-character lines. Acceptable, but if you want the printer to behave well at narrow widths in the future, this would need to be emitted as composite tokens (`pp_open_block(…, PP_OPEN_PAR)` / `pp_string` / `pp_close_block`) rather than a flat string. Not a blocker.

**7. Test brittleness.**
`algebraic-model-print.smt2.gold` and the updated `incremental00.smt2.gold` / `model-eval.smt2.gold` / `issue311{a,b}.smt2.gold` pin specific isolating intervals like `(/ 45.0 32.0)`. If libpoly's interval refinement ever tightens (a perfectly valid, internal change), every one of these gold files breaks even though the math output is still correct. Not introduced by this PR — the previous gold files pinned `1.414214`, which was equally fragile — but since the new format exposes more libpoly internals (degree, intermediate refinement), the surface area is bigger.

If you want to harden these tests later, an option is a python-side checker that parses `(root-of-with-interval …)` and verifies the *value* (substitute candidate root into polynomial, check sign, check bounds) instead of bytewise gold compare. Out of scope for this PR.

## Minor style

- `safe_malloc((deg + 1) * sizeof(lp_integer_t))` then a loop of `lp_integer_construct` — fine. Could note that this is parallel to libpoly's own `lp_upolynomial_unpack` precondition.
- `bool sqfree_allocated` — clear naming, good.
- The `append_lp_integer_smt2_coeff`/`append_dyadic_rational_smt2_real` helpers are clean and reusable. If a similar emitter is later needed elsewhere (e.g. for `(define-fun)` model output), worth promoting them to a small `smt2_lp_print.c` rather than keeping them `static` — but only if/when a second caller appears. YAGNI for now.
- Typo-grade: header comment block on `append_dyadic_rational_smt2_real` says `q = q->a / 2^(q->n)`. Convention in libpoly is `q.a / 2^q.n` since `q` is a value, not a pointer in the math; a tiny clarification.

## Tests

- `algebraic-model-print.smt2` (proper √2) and `algebraic-rational-print.smt2` (rational point, exercises `f == NULL`) cover both branches. ✅
- The four updated gold files transitively re-validate the printer on existing benchmarks.
- Missing: a test where the *input* polynomial has a non-trivial square (so the square-free reduction actually fires and changes the output), and one where the *primitive part* normalization fires (e.g. polynomial with content > 1). Both branches are exercised by ordinary roots, but a targeted test would document intent. Optional.

## Verdict

**LGTM with small revisions.** Top of my list, in order of importance:

1. Comment or assert on the open-interval assumption (point 1).
2. Add a `RELEASE_NOTES` / changelog line about the output format change (point 5).
3. Cite the SMT-COMP 2026 spec inline (point 4).
4. `size_t → unsigned long` cast (point 2).

The rest are nits or follow-ups.

disteph:
New code review by Windsurf/Opus4.7

# Final review — PR #623 (commit `3e1395f`)

## What `3e1395f` addresses (vs. my prior review)

| # | Prior issue | Status | Notes |
|---|---|---|---|
| 1 | Open/closed interval bits silently dropped | ✅ Fixed | `assert(an->I.a_open && an->I.b_open)` placed at the correct spot — *after* the `an->f == NULL` rational-point fast path, where the assumption actually applies (rational points have `a_open == b_open == 0` by libpoly convention, so a top-level assert would have fired spuriously). Good scoping. |
| 2 | `q->n` (`size_t`) → `unsigned long` truncation on LLP64 | ✅ Fixed | Explicit `(unsigned long)q->n` cast with a one-line justification comment. Acceptable. |
| 4 | No in-tree pointer to the SMT-COMP 2026 spec | ✅ Fixed | New 4-line header comment above `smt2_pp_algebraic` citing both `https://smt-comp.github.io/2026/` and PR #623. Future archaeology problem solved. |

## What is still open

| # | Prior issue | Status | Recommendation |
|---|---|---|---|
| 3 | `free()` matches GMP default allocator only | Not addressed | **Optional**, very minor. A one-line comment near `free(num_str)` / `free(den_str)` / `free(s)` would be a five-second fix, but Yices's existing call sites (e.g. `string_buffer_append_mpz`) follow the same convention without comment, so this is fine to defer. Don't block on this. |
| 5 | Backward-incompatible output for `(get-value)` / `(get-model)` | Not addressed | **Worth flagging before merge** — at minimum a `RELEASE_NOTES` / `NEWS` entry. The gold-file diffs in the PR (`tests/regress/mcsat/nra/incremental/incremental00.smt2.gold`, `model-eval.smt2.gold`, `wd/issue311a.smt2.gold`, `wd/issue311b.smt2.gold`) make the format change concrete: any downstream consumer parsing decimal output will break silently. Whether to add an opt-out flag is a separate product decision that can be deferred to a follow-up PR. |
| 6 | No pretty-printer line wrapping inside `(root-of-with-interval …)` | Explicit non-blocker | Leave as-is. |

## New, post-`3e1395f` observations

I re-read the merged-head version of `src/frontend/smt2/smt2_printer.c` (lines 208-289) and its companions `append_lp_integer_smt2_coeff` / `append_dyadic_rational_smt2_real` carefully. No regressions introduced by the review-fix commit. Three small stylistic notes — none blocking:

- **Assertion placement is intentional.** Worth a one-line comment if you want to discourage future "lift the assert to the top of the function" refactors: e.g. `/* Only meaningful for the proper-algebraic branch; rational points use closed [a, a]. */` next to the assert. Optional.
- **Spec link granularity.** `https://smt-comp.github.io/2026/` is a top-level page; if SMT-COMP publishes a stable anchor for the model-validation format (e.g. `…/specs.html#model-validation`), prefer that. Not blocking — the PR # in the same comment is the durable backstop.
- **Test coverage looks right.** The two new regressions cover both branches:
  - `tests/regress/mcsat/nra/algebraic-model-print.smt2` — irrational case (`x² = 2`), exercises the `(root-of-with-interval (coeffs ...) min max)` path *and* the new dyadic-real printer (`/ 45.0 32.0`, `/ 23.0 16.0`).
  - `tests/regress/mcsat/nra/algebraic-rational-print.smt2` — rational-point fast path (`x² = 4` ⇒ `2.0`), exercises `an->f == NULL`.

  Plus three updated gold files for previously-decimal output. Good coverage.

## Verdict

**Approve.** The three items I asked for are addressed cleanly and at the right scope. Recommend the author add a one-line `RELEASE_NOTES` / `NEWS` bullet about the output-format change (item 5) before merging, since that's the only thing with downstream impact; everything else can land as-is.

Issue #615, opened 2026-03-13
Yices crashes during interpolation


daniel-raffler:
Hello,

We've seen some occasional crashes during interpolation. The problem appears to be deterministic, however, I've been having a hard time finding a minimal example that will trigger the crash. What I can provide for now is a stacktrace:

```
Stack: [0x00007106186cc000,0x00007106187cc000],  sp=0x00007106187c9400,  free space=1013k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libyices2java.so+0x17b787]  mcsat_decide_assumption+0x403
C  [libyices2java.so+0x17c996]  mcsat_solve+0x487
C  [libyices2java.so+0x59efb]  check_context_with_model+0x5a
C  [libyices2java.so+0x37cad]  _o_yices_check_context_with_model+0x1c9
C  [libyices2java.so+0x37d4e]  yices_check_context_with_model+0x54
C  [libyices2java.so+0x3827a]  yices_check_context_with_interpolation+0x160
C  [libyices2java.so+0x1e982]  Java_com_sri_yices_Yices_checkContextWithInterpolation+0x102
j  com.sri.yices.Yices.checkContextWithInterpolation(JJJ[J[I)I+0
j  com.sri.yices.InterpolationContext.check(Lcom/sri/yices/Parameters;Z)Lcom/sri/yices/Status;+53
```
When assertions are enabled, this message is printed instead:
```
mcsat/value.c:478: mcsat_value_construct_from_value: Assertion `false' failed.
```

I've been using `issue-613-mcsat-curried-function-model` for my tests as I initially thought the problem may be related. However, it still seems to happen, even with the new branch

Do you have an idea what could be causing the issue?

disteph:
Resolved in #624

PR #624, opened 2026-05-05
Fix issue #615: crash in mcsat_value_construct_from_value during interpolation.

disteph:
When yices_check_context_with_model,
yices_check_context_with_model_and_hint, or
yices_check_context_with_interpolation was called with assumption variables whose type kind MCSAT cannot decide on (uninterpreted, function, tuple, or finite-field types), Yices crashed with

  mcsat/value.c:478: mcsat_value_construct_from_value:
    Assertion `false' failed.

(or a SIGSEGV in release builds), inside the call stack:

  mcsat_decide_assumption -> mcsat_value_construct_from_value(<unsupported>)

Root cause:

* mcsat_value_construct_from_value handles only BOOLEAN, RATIONAL, ALGEBRAIC, FINITEFIELD, and BITVECTOR value kinds; it asserts(false) on every other kind.

* Independently, the uf_plugin (decision-maker for UNINTERPRETED_TYPE and FUNCTION_TYPE) and ff_plugin (decision-maker for FF_TYPE) have decide_assignment == NULL, so even if value construction succeeded the next step in mcsat_decide_assumption would dereference NULL.

* The API entry points only verified that the assumption term was uninterpreted, not that its type was one MCSAT could actually decide on, so unsupported types reached the solver and tripped the assertion / crashed.

Fix: validate the type kind of every assumption term at the API boundary and reject anything that is not BOOL_TYPE, INT_TYPE, REAL_TYPE, SCALAR_TYPE, or BITVECTOR_TYPE (the type kinds whose plugins provide a non-NULL decide_assignment).

A new error code MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED distinguishes "wrong type" (this fix) from the existing MCSAT_ERROR_ASSUMPTION_TERM_NOT_SUPPORTED ("not an uninterpreted term"). Both yices_check_context_with_model and
yices_check_context_with_model_and_hint now report the new code, and yices_check_context_with_interpolation propagates it through its internal yices_check_context_with_model call instead of crashing.

Files:

* src/api/yices_api.c            -- new check_mcsat_assumption_types
                                    helper plus _o_good_assumption_terms_for_mcsat;
                                    wired into the two check_with_model
                                    entry points.
* src/include/yices_types.h      -- new MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED.
* src/include/yices.h            -- documented the new error code on
                                    the affected functions.
* src/api/yices_error.c          -- yices_print_error / yices_error_string.
* src/frontend/smt2/smt2_commands.c
                                 -- SMT2 frontend error printer.
* tests/api/test_issue_615_assumption_type.c
                                 -- regression test covering uninterpreted-sort,
                                    function-type, supported-type, and
                                    non-uninterpreted-term cases.

Verified:

* New regression test passes in debug and release.
* make MODE=release check       1705 pass / 0 fail.
* make MODE=release check-api   31/31 pass.


disteph:
Commit [1bf86b8](https://github.com/SRI-CSL/yices2/pull/624/commits/1bf86b892b08b4f604b44e969bf43165f45ef56f) extends the issue 615 fix from “reject unsupported model assumptions before they crash MCSAT” to “support tuple-valued model assumptions when they can be safely reduced to existing MCSAT decision types.”

The implemented plan was to keep function-valued tuple leaves out of scope, but allow tuples whose recursively flattened leaves are Bool, Int, Real, scalar, or BitVector. The API validation now accepts exactly those tuple types and still rejects tuples containing function, uninterpreted-sort, finite-field, or otherwise unsupported leaves with `MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED`.

Internally, MCSAT now tuple-blasts tuple assumptions, recursively flattens the corresponding tuple value from the input model, and decides/caches each scalar leaf value instead of trying to convert the aggregate tuple value into a single `mcsat_value_t`. Model hints use the same flattening path. The interpolation cleanup path was also fixed so errors propagated from internal model refutation do not leave the two interpolation contexts pushed.

The post-development conclusion is that tuple-valued input-model assumptions are now supported for the same scalar leaf types MCSAT already knows how to decide, without adding tuple support to `mcsat_value_construct_from_value`. Function-valued tuple leaves remain a separate, larger problem as planned.

Verification passed:
- `make -s check-api`
- `make -s MODE=debug check-api`
- release/debug `test_issue_615_assumption_type`
- `git diff --check`

disteph:
Code review provided by Windsurf/Opus4.7:

# Code review — `b53a1406` ("mcsat: support tuple-valued model assumptions")

## Summary

This commit closes issue #615 (MCSAT crashes on model assumptions of unsupported types) and goes one step further by promoting `TUPLE_TYPE` from "rejected" to "fully supported" whenever the recursively flattened tuple leaves are all MCSAT-decidable scalars (Bool, Int, Real, scalar, BitVector). Anything else — uninterpreted-sort, function, finite-field, function-typed tuple leaves — still fails cleanly at the API boundary with the new `MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED` error code instead of asserting / segfaulting.

Diff against master: 9 files, ~792/47 +/-, the bulk of which is the new test file.

Touched surface:
- API guard and error plumbing ([src/api/yices_api.c](cci:7://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:0:0-0:0), `src/api/yices_error.c`, `src/include/yices.h`, `src/include/yices_types.h`, `src/frontend/smt2/smt2_commands.c`).
- Preprocessor: a new public symbol [preprocessor_tuple_blast](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:10:0-13:1) exposing the tuple-blast pipeline already used internally for assertions.
- MCSAT solver: a parallel `assumption_values` vector, a tuple-aware decision/hint registration path, and a SCALAR-aware value constructor.
- A new test file [tests/api/test_issue_615_assumption_type.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:0:0-0:0) with 13 test cases.

## Strengths

- **Properly recursive API guard** in [mcsat_assumption_type_supported](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:2536:0-2561:1) at `src/api/yices_api.c:2537-2563`. Tuple-of-tuple is allowed iff every leaf is one of the five decidable kinds; everything else falls through to `default: return false`. The check is run in the existing per-term validation loop (`:2579-2587`) and sets `error->term1/type1` so downstream callers can pinpoint the offending term. Doc updates in `src/include/yices.h:3373-3375` and `:3426-3428` match.

- **Error-code design distinguishes "wrong shape" from "wrong type".** The previous shape check (`VARIABLE_REQUIRED` → `MCSAT_ERROR_ASSUMPTION_TERM_NOT_SUPPORTED`) is preserved unchanged; the new type check raises `MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED`, with an explanatory comment at `src/api/yices_api.c:9659-9667`. That separation is useful for callers and for the SMT2 frontend's pretty-printing.

- **Parallel `assumption_values` vector** in the solver. Pre-fix, [mcsat_decide_assumption](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:9:0-20:19) looked up `model_get_term_value(mdl, var_term)` at decision time, which only works when the assumption variable is a user-visible term that lives in the model. Tuple-blasted leaves are synthetic `new_uninterpreted_term`s the user has never seen, so that lookup would fail. Carrying the leaf value alongside the leaf variable at registration time ([mcsat_add_assumption_leaf](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2805:0-2811:1) at `src/mcsat/solver.c:2807-2812`) is the correct data shape and keeps [mcsat_decide_assumption](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:9:0-20:19) a pure consumer.

- **Lifecycle is symmetric**: `assumption_values` is initialized in [mcsat_construct](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:867:0-974:1), deleted in [mcsat_destruct](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:976:0-1014:1), reset both in [mcsat_clear](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:6:0-16:1) and at `solve_done:` in [mcsat_solve](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2941:0-3128:1), and an `assert(mcsat->assumption_values.size == 0)` at solve start mirrors the longstanding `assumption_vars` invariant. No init-once/leak/use-after-reset paths visible.

- **[mcsat_decide_assumption](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:9:0-20:19) signature is properly narrowed** to `value_table_t* vtbl` rather than a full `model_t*` (`src/mcsat/solver.c:2461`). The body really only ever needed the value table, and the single caller in [mcsat_solve](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2941:0-3128:1) passes `model_get_vtbl(mdl)`. Minor but correct.

- **Shared blast+flatten helper** [mcsat_collect_tuple_leaves_and_values](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2813:0-2822:1) at `src/mcsat/solver.c:2814-2823` factors the common pre-amble out of [mcsat_add_tuple_assumption_leaves](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2824:0-2854:1) and [mcsat_set_tuple_hint_leaves](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2866:0-2892:1). The remaining per-caller bodies are short and intentionally different.

- **Hint-vs-assumption asymmetry is documented in code.** The assumption path runs [preprocessor_apply](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:4:0-30:2) per leaf and re-asserts `leaf = leaf_pre` (`src/mcsat/solver.c:2841-2851`); the hint path does not, and the comment at `:2884-2886` calls that out:

  > Hints are advisory cache entries, not assumption decisions. No preprocessor_apply/equality assertion is needed here: if a leaf was substituted, search can ignore or overwrite the stale cached hint.

  Decision recorded.

- **SCALAR embedding documented inline** (`src/mcsat/solver.c:2160-2175`):

  > Scalar values are decided by the rational plugin; the scalar constant index is the corresponding integer value.

  This matters because `mcsat_value_construct_from_value` only handles `BOOLEAN/RATIONAL/ALGEBRAIC/FINITEFIELD/BITVECTOR`; SCALAR values would otherwise fall through to its `assert(false)` arm. The wrapper [mcsat_value_construct_from_typed_model_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2158:0-2174:1) does the SCALAR conversion explicitly and delegates everything else.

- **Interpolation cleanup is now correct on error.** [yices_check_context_with_interpolation](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9854:0-9991:1) (`src/api/yices_api.c:9959-9982`) always pops both pushed contexts, even when popping `ctx_B` fails, and saves+restores the original error report so an inner error code (e.g. `MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED` raised mid-loop) survives the cleanup. Strict improvement: pre-fix the contexts could be left unbalanced on error and the inner error code was not stable across pop failures.

- **Defensive failure paths are annotated.** Both tuple registrations longjmp on a malformed user model, with a one-line comment

  > Defensive path for malformed user models: API validation has already rejected unsupported tuple leaf types before solver entry.

  pinning the actual trigger (the user gave us a model whose value for the tuple is the wrong shape, or absent).

- **[preprocessor_tuple_blast](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:10:0-13:1) docstring spells out the contract** at `src/mcsat/preprocessor.h:111-117`: `out` is reset by the callee, leaf order matches [type_collect_blasted_atom_types](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:134:0-147:1), leaves are memoized and stable across later calls. The implementation in [src/mcsat/preprocessor.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:0:0-0:0) `ivector_reset(out)`s explicitly even though [tuple_blast_get](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:169:0-177:1) already resets, which makes the public contract self-evident at the entry point (belt-and-suspenders, harmless).

## Concerns

### Minor

1. **[mcsat_collect_tuple_leaves_and_values](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2813:0-2822:1) collapses two failure modes into one bool.** The helper returns `false` for either [mcsat_flatten_model_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2126:0-2156:1) failure or `leaves->size != values->size`. Both callers treat them identically and longjmp with `MCSAT_EXCEPTION_UNSUPPORTED_THEORY`. That's fine today, but `MCSAT_EXCEPTION_UNSUPPORTED_THEORY` is the wrong code for "user model has the wrong shape" — the theory *is* supported, the input is malformed. If a future change adds API-boundary validation that the model has a value of the right shape for each tuple assumption, this longjmp becomes genuinely unreachable and either the helper can be simplified (returning `void` plus `assert`) or the failure code can be specialized. Not blocking.

2. **[mcsat_collect_tuple_leaves_and_values](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2813:0-2822:1) expects the out-param ivectors to be empty at entry.** The post-condition `leaves->size == values->size` is only meaningful relative to the in-bound state. Both current callers init immediately before the call so the assumption holds, but if a third caller ever appears a one-line `assert(leaves->size == 0 && values->size == 0)` at entry would be cheap insurance.

3. **[mcsat_add_tuple_assumption_leaves](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2824:0-2854:1) runs [preprocessor_apply](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:4:0-30:2) per leaf** (`src/mcsat/solver.c:2843`), which is the right call for soundness when a leaf has been substituted by an earlier assertion. But the loop calls [preprocessor_apply](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:4:0-30:2) even for leaves that have never been touched — every call hits the `preprocess_map` hash and the `preprocess_map` is shared across all assertions. For a tuple of arity *k* with *n* assumption variables this is `O(n·k)` hash lookups per solve. Not a problem in practice (k is small), but if you ever have wide tuples it might be cheap to skip the call when the leaf was minted in this solve and could not have been substituted. Optional.

4. **[build_value_from_flat](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:12:0-37:1) and [merge_blasted_function_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:86:0-224:1) in the preprocessor (called indirectly via [preprocessor_build_tuple_model](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:226:0-322:1))** depend on [mcsat_flatten_model_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2126:0-2156:1)'s recursion order matching [type_collect_flat](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:85:0-106:1)'s. That invariant is currently true by construction but is enforced only by the tests. A doc comment on [mcsat_flatten_model_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2126:0-2156:1) ("recursion order MUST match [type_collect_flat](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:85:0-106:1)") would harden this against future drift. Not blocking.

## Test coverage

[tests/api/test_issue_615_assumption_type.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:0:0-0:0) runs a focused suite under [yices_has_mcsat()](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:1290:0-1292:1):

**Rejection paths**
- `check_uninterpreted_sort_assumption_rejected` — uninterpreted-sort assumption variables, both `_with_model` and `_with_model_and_hint`. Reproduces the original issue #615 crash trace.
- `check_function_type_assumption_rejected` — function-typed assumption variable.
- `check_tuple_with_unsupported_leaf_rejected` — three tuple-of-{uninterp, function, FF} variants, each must produce `MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED`.
- `check_term_shape_check_preserved` — non-uninterpreted assumption term still yields the older `MCSAT_ERROR_ASSUMPTION_TERM_NOT_SUPPORTED` (regression guard for the shape vs. type distinction).

**Supported scalar paths**
- `check_supported_types_still_work` — Real and SCALAR(3) assumption variables both reach `YICES_STATUS_SAT/UNKNOWN`. The SCALAR sub-case actually exercises the new [mcsat_value_construct_from_typed_model_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2158:0-2174:1) SCALAR arm.

**Supported tuple paths**
- [check_tuple_assumption_sat](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:32:0-38:32), `check_tuple_assumption_unsat`, [check_nested_tuple_assumption_sat](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:31:0-40:32) — flat and nested tuples, model-consistent and model-inconsistent.
- [check_tuple_with_scalar_leaf](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:10:0-31:1) — cross-product `(int, scalar(3))` exercising both the tuple-flatten path and the SCALAR-via-rational construction together.
- [check_tuple_assumption_respects_preprocessing](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:10:0-29:1) — asserts `x = (2, true)` and then runs `_with_model` with the model fixing `x` to `(3, true)`, expecting UNSAT. Targeted regression for the per-leaf [preprocessor_apply](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:4:0-30:2) call: without it, leaf-level substitution would hide the contradiction.
- [check_tuple_repeat_solve_resets_assumptions](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:33:0-54:1) — calls `_with_model` twice on the same context with a tuple assumption. Pins the `assumption_values` reset between solves; without the reset, the `assert(mcsat->assumption_values.size == 0)` at solve start would fire on the second call.
- [check_tuple_model_hint](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:56:0-65:98) — tuple-typed hint via `_with_model_and_hint`.
- `check_tuple_interpolation_refutation` — tuple-typed assumption inside [yices_check_context_with_interpolation](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9854:0-9991:1), exercises the cleanup path on the UNSAT branch.

**Coverage assessment**

The suite is well-targeted: each non-trivial code path in the diff has at least one test that would fail if that path regressed. The two paths I'd still call slightly under-covered are:
- *the longjmp arm in [mcsat_collect_tuple_leaves_and_values](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:2813:0-2822:1)*: there is no test that calls `_with_model` with a tuple assumption variable but a model that omits or shape-mismatches that variable. A "user supplies an empty model for a tuple assumption" test would either prove the longjmp fires cleanly or motivate moving the check to the API boundary (concern #1 above).
- *interpolation error preservation*: `check_tuple_interpolation_refutation` covers the success path; there is no test that surfaces an internal error (e.g. unsupported type) through [yices_check_context_with_interpolation](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9854:0-9991:1) and asserts the error code survives the cleanup pops. The preservation logic at `src/api/yices_api.c:9959-9982` is untested.

Neither gap is a merge-blocker.

## Verdict

LGTM. The design is sound, the lifecycle accounting is symmetric, the hint/assumption asymmetry is intentional and called out, the SCALAR sub-fix is documented and tested, and the test suite covers each non-trivial new code path.

The remaining items (helper-failure-mode granularity, redundant [preprocessor_apply](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:4:0-30:2) calls on virgin leaves, doc comments on flatten ordering, the two coverage gaps) are all follow-up tier — none of them block the merge.

disteph:
Update on the coverage assessment after 2139806:

Both gaps I'd flagged are now closed.

- **[mcsat_collect_tuple_leaves_and_values](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:18:0-27:1) longjmp / "model omits the tuple assumption" path** — covered by [check_tuple_empty_model_uses_default_value](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:338:0-354:1) in [tests/api/test_issue_615_assumption_type.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:0:0-0:0). It hands `yices_check_context_with_model` an empty `model_t` together with a tuple-typed assumption variable, and asserts the call returns cleanly (`SAT` or `UNKNOWN`, no `YICES_NO_ERROR` violation). That exercises the "tuple leaves are unmapped, fall through to defaults" branch end-to-end without going through the longjmp.
- **Interpolation error preservation across the cleanup pops** — covered by [check_interpolation_preserves_refutation_error](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_issue_615_assumption_type.c:401:0-422:1). `ctx_A` is mcsat-with-interpolation, `ctx_B` is a default context asserting an inequality over an uninterpreted-sort variable. The internal model refutation in `ctx_A` then runs `_with_model` over an unsupported assumption type, raising `MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED`. The test asserts that this error code survives the cleanup-pops in `yices_check_context_with_interpolation` (`src/api/yices_api.c:9959-9982`). That's the exact preservation path I called untested.

With these in, every non-trivial branch in the diff has at least one regression test that would fail if it regressed. No remaining coverage concerns from my side.

PR #628, opened 2026-05-08
mcsat: per-cnf gc-mark progress index (issue #616)

disteph:
Tentative fix of #616. Original bug could not be reproduced, so whether this addresses the issue is guesswork.

# Issue #616 — [mcsat_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1128:0-1200:1) GC paths had two distinct thread-/instance-safety bugs

The reporter's stack trace is a deterministic signature:

```
good_term_idx
good_term
variable_db_is_variable
clause_check
clause_db_is_clause
clause_db_gc_sweep
bool_plugin_gc_sweep
mcsat_gc
mcsat_pop
context_pop
yices_pop
```

`clause_db_gc_sweep` reaches a clause that should still be live, dereferences its literals' variable terms, and `good_term` rejects them — i.e. the GC sweep freed something that the GC mark phase failed to mark.

Auditing the MCSAT push/pop/GC paths surfaced two independent hazards on this exact call chain. This PR fixes both, one per commit (not squashed).

### Root cause

[cnf_gc_mark](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/bool/cnf.c:391:0-423:1) in [src/mcsat/bool/cnf.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/bool/cnf.c:0:0-0:0) tracked its progress through `gc_vars->marked` with a function-local `static uint32_t i`:

```c
void cnf_gc_mark(cnf_t* cnf, gc_info_t* gc_clauses, const gc_info_t* gc_vars) {
  static uint32_t i;
  ...
  if (gc_vars->level == 0) { i = 0; }
  for (; i < gc_vars->marked.size; ++ i) { ... }
}
```

The `static` is one object in `.bss`, shared by every `cnf_t` instance in the process — but the surrounding code reads and writes it as if it belonged to the `cnf_t` argument (reset only at `gc_vars->level == 0`, otherwise resumed across calls within one GC round). That assumption breaks as soon as more than one `cnf_t`'s mark phase runs against the same `static i`.

`gc_info_mark` is idempotent, so re-marking is harmless. The damage comes from the opposite direction: when the index is non-zero on entry because a *different* `cnf_t` left it there, the loop starts at offset `i` and **skips the prefix `[0, i)` of its own `gc_vars->marked` array**. Clauses defining the skipped variables are never marked, then `clause_db_gc_sweep` frees them, and the next walk over the clause DB dereferences freed memory inside `clause_db_is_clause → variable_db_is_variable → good_term`. That matches the reporter's stack trace.

### How the bad interleaving arises

The hazard is intrinsically multi-threaded. A single thread cannot trigger it: [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) walks contexts sequentially and constructs a fresh `gc_vars` per context, so each context's first [cnf_gc_mark](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/bool/cnf.c:391:0-423:1) call sees `gc_vars->level == 0` and resets the static.

Under threads, two [yices_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9103:0-9109:1) calls on different MCSAT contexts both reach `mcsat_pop → mcsat_gc(false) → bool_plugin_gc_mark → cnf_gc_mark` and both write the same `static i`. With `libyices2java.so` and a JVM dispatching, two MCSAT contexts in two threads is a routine pattern.

### Fix

Move the index from a function-local `static` to a per-instance field on `cnf_t`:

- `src/mcsat/bool/cnf.h` — add `uint32_t gc_mark_index;` to `struct cnf_s`.
- [src/mcsat/bool/cnf.c::cnf_construct](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/bool/cnf.c:22:0-28:1) — initialise to 0.
- [src/mcsat/bool/cnf.c::cnf_gc_mark](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/bool/cnf.c:391:0-423:1) — replace every `i` with `cnf->gc_mark_index`.

A `uint32_t i = 0;` stack-local would also have fixed correctness (marking is idempotent) but at the cost of redundantly re-walking the already-processed prefix on every level of the fixed-point loop. The per-instance field preserves the original O(N) behaviour the static was there to provide.

### Root cause

[yices_push](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9017:0-9045:1) and [yices_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9103:0-9109:1) in [src/api/yices_api.c](cci:7://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:0:0-0:0) were not lock-protected at all. On an MCSAT-enabled context this is unsound: both APIs mutate per-context MCSAT state (trail, plugin internal state, scope holder, preprocessor) via [context_clear](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context.c:6621:0-6634:1) / [context_clear_unsat](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context.c:6638:0-6658:1) / [context_push](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context.c:5944:0-5960:1) / [context_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context.c:5962:0-5977:1).

[yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) holds `__yices_globals.lock` and walks every live context via `context_list_gc_mark → context_gc_mark`, which mutates exactly the same per-context state: it resets the `ctx->top_* / subst_eqs / aux_eqs` vectors, calls `intern_tbl_gc_mark` / `egraph_gc_mark` / `fun_solver_gc_mark`, and on MCSAT contexts runs the full [mcsat_gc(true)](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1299:0-1446:1) (mark + sweep on the trail and every plugin).

For MCSAT contexts, [mcsat_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1128:0-1200:1) additionally reads `__yices_globals.terms` via `variable_db_get_term` / `term_type`, racing against `term_table_gc`.

So the unprotected push and pop race against:

- [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) (on per-context MCSAT state, plus the global term table for pop),
- [yices_assert_formula](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9222:0-9225:1) and other term-construction APIs (mutate the global term table that pop reads),
- a second thread inside [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) that, under its own held lock, calls into [mcsat_gc(true)](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1299:0-1446:1) on the same MCSAT context — i.e. two threads inside [mcsat_gc](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1299:0-1446:1) on the same `mcsat_solver_t`.

### Fix

Split the bodies into [_o_yices_push](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:8966:0-9015:1) / [_o_yices_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:9049:0-9101:1) (the original bodies, unchanged) plus thin dispatchers that take `__yices_globals.lock` only on the MCSAT branch:

```c
EXPORTED int32_t yices_push(context_t *ctx) {
  if (ctx->mcsat != NULL) {
    MT_PROTECT(int32_t, __yices_globals.lock, _o_yices_push(ctx));
  }
  return _o_yices_push(ctx);
}

EXPORTED int32_t yices_pop(context_t *ctx) {
  if (ctx->mcsat != NULL) {
    MT_PROTECT(int32_t, __yices_globals.lock, _o_yices_pop(ctx));
  }
  return _o_yices_pop(ctx);
}
```

CDCL(T) push and pop continue to run lock-free.

This is the same conditional-locking pattern already in use one level down in [src/context/context_solver.c](cci:7://file:///Users/sgl/git/yices/yices2/src/context/context_solver.c:0:0-0:0): [call_mcsat_solver](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context_solver.c:563:0-565:1) and [check_context_with_term_assumptions_mcsat](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context_solver.c:1332:0-1334:1) are `MT_PROTECT`'d, while the CDCL(T) [solve()](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context_solver.c:371:0-472:1) body in [check_context](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context_solver.c:567:0-594:1) and [check_context_with_assumptions](cci:1://file:///Users/sgl/git/yices/yices2/src/context/context_solver.c:597:0-620:1) is not.

Reading `ctx->mcsat` without the lock is safe: it is set once at context construction and never mutated afterwards.

## Combined effect

Together the two commits close three closely related races on the reporter's `yices_pop → mcsat_pop → mcsat_gc` stack:

- Cross-`cnf_t` race on the `static i` in [cnf_gc_mark](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/bool/cnf.c:391:0-423:1) (commit 1).
- [mcsat_gc(false)](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1299:0-1446:1) vs. [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) / term construction on `__yices_globals.terms` (commit 2).
- [mcsat_clear](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1202:0-1212:1) / [mcsat_push](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1092:0-1125:1) / [mcsat_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1128:0-1200:1) vs. [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) on per-context MCSAT state — including the more subtle "two threads inside [mcsat_gc](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1299:0-1446:1) on the same context via push/pop-vs-[yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1)" interleaving (commit 2).

I was not able to reproduce the original `libyices2java.so` crash, so I cannot prove these are the trigger. Each commit is positioned as a minimal upstream fix that removes a concrete, demonstrable hazard on the reported call chain.

## Test

`tests/api/test_mcsat_pop_gc_multi_context.c` (commit 1) — a single-threaded **smoke test** of the GC fan-out path: it builds three live MCSAT contexts, asserts CNF-flavoured Boolean formulas in each, pushes and asserts more lemmas, calls [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) (which walks every context's [mcsat_gc_mark](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:3293:0-3296:1)), then pops on every context (driving `mcsat_pop → mcsat_gc(false) → bool_plugin_gc_sweep → clause_db_gc_sweep`).

This test does **not** deterministically reproduce the cross-`cnf_t` static-i bug — that bug requires concurrency, since [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) runs each context's [mcsat_gc(true)](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1299:0-1446:1) to completion synchronously and constructs a fresh `gc_vars` per context, resetting the static at every context boundary. The test is included as end-to-end coverage of the multi-context GC fan-out, which was previously underexercised. A targeted multi-threaded stress test belongs in a separate PR alongside an MT stress harness.

The push/pop-vs-[yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) races fixed by commit 2 are also intrinsically multi-threaded and not covered by a deterministic regression test in this PR; the existing single-threaded suite plus `--enable-thread-safety` builds cover the macro expansion and the dispatcher correctness.

## Verification

Reconfigured with `./configure --enable-thread-safety --enable-mcsat`, full clean `MODE=debug` rebuild, both commits applied:

- `make MODE=debug` builds cleanly.
- `make MODE=debug check-api` — **35 / 35 pass** (including `test_mcsat_pop_gc_multi_context`).
- `make MODE=debug mcsat-check` — **696 / 696 pass**.

If your usual configuration differs, re-run `./configure` with your flags before continuing.

## Out of scope (recommended follow-up)

- `src/mcsat/preprocessor.c::composite_for_noncomposite` (around line 1242) is another function-local `static` on a non-GC path — same class of cross-instance hazard, lower severity.
- The CDCL(T) push/pop paths still race against [yices_garbage_collect](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13469:0-13482:1) on per-context state (`intern_tbl_push` vs. `intern_tbl_gc_mark`, etc.). This is a longstanding upstream issue, unrelated to MCSAT or to issue #616, and orthogonal to this PR's scope. Worth tracking separately.

PR supported by Windsurf/Opus4.7.

### DIFF
--- a/src/api/yices_api.c
+++ b/src/api/yices_api.c
@@ -8975,7 +8975,7 @@ EXPORTED void yices_reset_context(context_t *ctx) {
  * - if the context status is UNSAT or SEARCHING or INTERRUPTED
  *   code = CTX_INVALID_OPERATION
  */
-EXPORTED int32_t yices_push(context_t *ctx) {
+static int32_t _o_yices_push(context_t *ctx) {
   if (! context_supports_pushpop(ctx)) {
     set_error_code(CTX_OPERATION_NOT_SUPPORTED);
     return -1;
@@ -9015,6 +9015,36 @@ EXPORTED int32_t yices_push(context_t *ctx) {
   return 0;
 }
 
+/*
+ * yices_push and yices_pop both mutate per-context MCSAT state
+ * (mcsat->trail, plugin internal state, scope holder, preprocessor)
+ * via context_clear / context_clear_unsat / context_push / context_pop.
+ *
+ * yices_garbage_collect holds __yices_globals.lock and walks every live
+ * context via context_list_gc_mark -> context_gc_mark, which mutates the
+ * same per-context state: it resets the ctx->top_* / subst_eqs / aux_eqs
+ * vectors, calls intern_tbl_gc_mark / egraph_gc_mark / fun_solver_gc_mark,
+ * and on MCSAT contexts runs the full mcsat_gc(true) (mark + sweep on the
+ * trail and every plugin).
+ *
+ * For MCSAT contexts, mcsat_pop additionally reads __yices_globals.terms
+ * via variable_db_get_term / term_type, racing against term_table_gc.
+ *
+ * Take __yices_globals.lock for the MCSAT branch only; CDCL(T) push/pop
+ * paths are left unlocked, matching the existing pattern in
+ * context_solver.c (call_mcsat_solver and check_context_with_term_assumptions_mcsat
+ * are MT_PROTECT'd; the CDCL(T) solve() body is not).
+ *
+ * Reading ctx->mcsat without the lock is safe: it is set once at context
+ * construction and never mutated afterwards.
+ */
+EXPORTED int32_t yices_push(context_t *ctx) {
+  if (ctx->mcsat != NULL) {
+    MT_PROTECT(int32_t, __yices_globals.lock, _o_yices_push(ctx));
+  }
+  return _o_yices_push(ctx);
+}
+
 
 
 /*
@@ -9029,7 +9059,7 @@ EXPORTED int32_t yices_push(context_t *ctx) {
  *   or if the context's status is SEARCHING or INTERRUPTED
  *   code = CTX_INVALID_OPERATION
  */
-EXPORTED int32_t yices_pop(context_t *ctx) {
+static int32_t _o_yices_pop(context_t *ctx) {
   if (! context_supports_pushpop(ctx)) {
     set_error_code(CTX_OPERATION_NOT_SUPPORTED);
     return -1;
@@ -9069,6 +9099,14 @@ EXPORTED int32_t yices_pop(context_t *ctx) {
   context_pop(ctx);
 
   return 0;
+}
+
+/* See the comment above yices_push for the locking rationale. */
+EXPORTED int32_t yices_pop(context_t *ctx) {
+  if (ctx->mcsat != NULL) {
+    MT_PROTECT(int32_t, __yices_globals.lock, _o_yices_pop(ctx));
+  }
+  return _o_yices_pop(ctx);
 }
 
 
--- a/src/mcsat/bool/cnf.c
+++ b/src/mcsat/bool/cnf.c
@@ -24,6 +24,7 @@ void cnf_construct(cnf_t* cnf, plugin_context_t* ctx, clause_db_t* clause_db) {
   cnf->clause_db = clause_db;
   cnf->ctx = ctx;
   cnf->variable = variable_null;
+  cnf->gc_mark_index = 0;
   int_lset_construct(&cnf->converted);
 }
 
@@ -389,20 +390,23 @@ void cnf_convert_lemma(cnf_t* cnf, const ivector_t* lemma, ivector_t* clauses) {
 }
 
 void cnf_gc_mark(cnf_t* cnf, gc_info_t* gc_clauses, const gc_info_t* gc_vars) {
-  static uint32_t i;
-
+  /* Per-cnf-instance progress index across the iterations of one mcsat_gc
+   * cycle. Used to be a function-local `static`, which (a) made the index
+   * shared across all cnf_t instances and (b) was a thread-safety concern
+   * under --enable-thread-safety. See cnf_t::gc_mark_index for details and
+   * issue #616. */
   variable_t var;
   clause_ref_t clause_ref;
   int_lset_iterator_t it;
 
   // If first time at gc, reset the index
   if (gc_vars->level == 0) {
-    i = 0;
+    cnf->gc_mark_index = 0;
   }
 
   // CNF marks only the clauses that are definitions of the variables to keep
-  for (; i < gc_vars->marked.size; ++ i) {
-    var = gc_vars->marked.data[i];
+  for (; cnf->gc_mark_index < gc_vars->marked.size; ++ cnf->gc_mark_index) {
+    var = gc_vars->marked.data[cnf->gc_mark_index];
     if (trace_enabled(cnf->ctx->tracer, "mcsat::gc")) {
       ctx_trace_term(cnf->ctx, variable_db_get_term(cnf->ctx->var_db, var));
     }
--- a/src/mcsat/bool/cnf.h
+++ b/src/mcsat/bool/cnf.h
@@ -45,6 +45,19 @@ typedef struct {
   /** Current variable being converted */
   variable_t variable;
 
+  /**
+   * Per-instance progress index for cnf_gc_mark across iterations of one
+   * mcsat_gc cycle. Reset to 0 when the cycle starts (gc_vars->level == 0)
+   * and otherwise advances across the for-each-plugin marking rounds.
+   *
+   * Was previously a function-local `static uint32_t` in cnf_gc_mark, which
+   * meant the index was shared across all live cnf_t instances. With one
+   * cnf_t per bool_plugin per context, that made the marking indices race
+   * across contexts under thread-safe builds and could in principle skip
+   * variables when two contexts' GC cycles interleaved (issue #616).
+   */
+  uint32_t gc_mark_index;
+
 } cnf_t;
 
 
--- a/tests/api/test_mcsat_pop_gc_multi_context.c
+++ b/tests/api/test_mcsat_pop_gc_multi_context.c
@@ -1,0 +1,205 @@
+/*
+ * Regression test exercising MCSAT pop + garbage collection across multiple
+ * live MCSAT contexts.
+ *
+ * Background (issue #616):
+ *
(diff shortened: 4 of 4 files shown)


ahmed-irfan (approved):

disteph:
Review provided by Codex/GPT5.5

**Findings**

No blocking findings.

**Residual Risk**

[tests/api/test_mcsat_pop_gc_multi_context.c](/Users/sgl/git/yices/yices2/tests/api/test_mcsat_pop_gc_multi_context.c:166) exercises the relevant MCSAT GC/pop path across multiple live contexts, but it does not deterministically reproduce either the old shared-static interleaving or the multi-threaded pop-vs-GC race. A targeted interleaving or MT stress test would still be useful follow-up coverage.

**Summary**

The `cnf_t::gc_mark_index` change correctly moves CNF GC progress from process-global static storage to per-instance state. The API locking change consistently protects MCSAT `yices_push` and `yices_pop` with `__yices_globals.lock`, while leaving CDCL(T) paths unchanged.

**Tests Run**

`git diff --check 6690298c80ac8426ded1115e27f802851e64ae02..HEAD`

`make MODE=debug check-api` passed: 35/35.

`make MODE=debug mcsat-check` passed: 696/696.

PR #630, opened 2026-05-18
Add wide (truth-invariant) model generalization mode

disteph:
# Add wide model generalization mode

## Motivation

`yices_generalize_model{,_array}` historically project a single literal
implicant of `F(X, y)` at the model, producing a sign-invariant cell
that is sound but unnecessarily narrow when `F` has top-level Boolean
structure satisfied by more than one disjunct. Refinement loops
(YicesQS-style) then re-enter the solver many times to exhaust the
formula one narrow cell at a time.

This PR adds an opt-in **wide** mode that enumerates multiple
model-true Boolean implicants of `F` via a SAT solver, projects each
through the legacy per-cube pipeline, and unions the results. The wide
cell is always at least as broad as the legacy cell and strictly
broader on many inputs with model-satisfied Boolean structure.

## Mode constants

| Mode | Algorithm |
|---|---|
| `YICES_GEN_BY_SUBST` | substitution (unchanged) |
| `YICES_GEN_BY_PROJ` | legacy implicant-then-project (unchanged) |
| `YICES_GEN_BY_PROJ_WIDE` | **new**: SAT-guided wide projection |
| `YICES_GEN_DEFAULT` | hybrid: substitution for discrete vars, `YICES_GEN_BY_PROJ` for real vars (unchanged) |

The wide path is reachable only via explicit `YICES_GEN_BY_PROJ_WIDE`
(or the new `_with_budget` entries below). Existing callers see no
behavioural change. The contract `G(X)` true at the model and
`G(X) ⇒ ∃y. F(X, y)` is preserved in all modes.

## Algorithm (wide)

`gen_model_by_proj_sat_guided`:

1. **Model-pruned, polarity-aware abstraction**: build a Boolean DAG
   of `F` against the model. Model-false subterms become `FALSE`
   constants and AND/OR simplify them away, so the SAT enumerator
   only ever sees the part of `F` the model satisfies.
2. **Tseitin clausification** into `new_sat_solver`; assert the root.
3. **SAT-guided enumeration loop**: `nsat_solve` → extract a monotone
   positive implicant → `project_one_cube_into` (the legacy
   `get_implicant` + `project_literals`) → block the implicant →
   repeat. Projection is **interleaved** with enumeration (per
   iteration, not after collecting cubes).
4. **Disjoin** the projected cubes. For purely conjunctive input the
   path short-circuits to the legacy pipeline and `v` is byte-identical
   to `YICES_GEN_BY_PROJ`.

Per-cube projection errors (e.g. unsupported term-kind in non-MCSAT
builds) are *skipped*: the failing implicant is blocked, enumeration
continues. Boolean enumeration is finite without a budget because every
iteration adds a blocker clause.

## New public APIs

```c
int32_t yices_generalize_model_with_budget(
    model_t *mdl, term_t t, uint32_t nelims, const term_t elim[],
    yices_gen_mode_t mode, uint32_t cube_budget, term_vector_t *v);

int32_t yices_generalize_model_array_with_budget(
    model_t *mdl, uint32_t n, const term_t a[], uint32_t nelims,
    const term_t elim[], yices_gen_mode_t mode, uint32_t cube_budget,
    term_vector_t *v);
```

`cube_budget` caps the number of SAT iterations (extracted cubes,
whether or not projection succeeds). `0` means unbounded. Attempts are
counted rather than successes so that a pathological `F` whose every
implicant uses an unsupported literal cannot drag the loop through all
`2^N` Boolean models. `cube_budget` is honoured only for
`YICES_GEN_BY_PROJ_WIDE`; it is ignored for the other modes.

The existing `yices_generalize_model{,_array}` are unchanged in
signature; they behave as `cube_budget = 0`.

## Failure handling

- **SAT engine returns non-`STAT_SAT` / non-`STAT_UNSAT`** → `assert`
  (documented impossible).
- **Precondition violations** (`F` not model-evaluable, or `F` false at
  the model) → respectively delegate to `gen_model_by_proj_local` (for
  its specific `GEN_EVAL_*` / `MDL_EVAL_*` code), or return
  `MDL_EVAL_FORMULA_FALSE` directly.
- **Per-cube projection failure** → skip, block, continue. Counts
  against `cube_budget`.
- **Budget hit with at least one successful cube** → return
  `OR(collected, gen_model_by_proj_local(F))` (broader than local).
- **No cube ever projected** → fall back to `gen_model_by_proj_local`
  for its error code.

## Tests

`tests/api/test_generalize_model.c`: 9 cases, every one checks
soundness (`G(X)` true at the model) and `AND(local) ⇒ AND(wide)`.
Highlights:

- `test_overlapping_arith_disjunction` (headline win, hard assert):
  eliminating `x` from `(y>x ∧ x>0) ∨ (z>x ∧ x>5)` against a model
  satisfying both disjuncts; wide must be strictly broader than local.
- `test_sat_guided_compression`: SAT-guided cube count `<` syntactic
  Cartesian product when atoms are shared.
- `test_sat_guided_budget_graceful`: calls `_with_budget` with
  `cube_budget = 4` against an input with up to `2^11` implicants;
  exercises the `OR(collected, local)` budget-exhaustion fallback.
- `test_sat_guided_polarity`: signed-literal abstraction + model-false
  pruning.

`make MODE=release check-api`: 36 pass, 0 fail.

## Diff stat

```
 src/api/yices_api.c               |   46 +-
 src/api/yices_api_lock_free.h     |    9 +
 src/include/yices.h               |   84 +-
 src/include/yices_types.h         |   27 +-
 src/model/generalization.c        | 1089 ++++++++++++++++++++++++++++
 src/model/generalization.h        |   55 +-
 tests/api/test_generalize_model.c |  716 +++++++++++++++++++
 7 files changed, 1981 insertions(+), 45 deletions(-)
```

## Backward compatibility

- **Source-level**: no signature change. Callers using
  `YICES_GEN_BY_PROJ` or `YICES_GEN_DEFAULT` get exactly their pre-PR
  behaviour.
- **ABI**: enum additively extended (`YICES_GEN_BY_PROJ_WIDE` appended);
  numeric values of existing entries unchanged.
- **Output shape**: under `YICES_GEN_BY_PROJ_WIDE`, multi-cube outputs
  return a single top-level disjunction in `v->data[0]` (single-cube
  outputs remain flat literals). All other modes keep the
  literal-per-element shape.

## Out of scope / follow-ups

- Per-disjunct model nudging (milestone 2, separate mode
  `YICES_GEN_BY_PROJ_WIDE_NUDGE`).
- `XOR_TERM` / iff decomposition (currently opaque leaves).
- Implicant minimisation.
- `efsolver` integration.
- Benchmark numbers on YicesQS / NRA-COMP — to be produced from the
  YicesQS side.

## Checklist

- [x] Builds clean in `MODE=debug` and `MODE=release`.
- [x] `make MODE=release check-api`: 36 pass.
- [x] New regression tests added.
- [x] Public header documented.
- [x] No changes outside the model-generalization dispatch.
- [x] Enum extension is additive.


disteph:
Review by Codex/GPT5.5:

Findings

No code findings. Reviewed feature/wide-model-projection against master; I did not find a correctness, API compatibility, dispatch, fallback, or test-integration issue in the branch diff.

Summary

The implementation matches the described opt-in behavior: YICES_GEN_BY_PROJ and YICES_GEN_DEFAULT keep legacy semantics, YICES_GEN_BY_PROJ_WIDE routes to the SAT-guided wide path, and cube_budget is honored only for wide mode. The SAT-guided loop reuses y2sat incrementally, blocks each extracted implicant, skips per-cube projection failures, and falls back to local projection when needed. The tests cover strict widening, shared-atom compression, parity, budget fallback, polarity, and shared elimination variables.

Tests Run

make MODE=release check-api

Result: 36 pass, 0 fail.

disteph:
# Code review by Windsurf/Opus4.7: SRI-CSL/yices2 PR #630 — Add wide (truth-invariant) model generalization mode

**Branch**: `feature/wide-model-projection`
**Head**: `8b09c89f` (12 commits ahead of master)
**Diff**: 7 files, 1971 insertions / 45 deletions
**Build / tests**: `make MODE={release,debug}` clean (no warnings); `make MODE=release check-api` 36/36 pass

---

## 1. What this PR does

It adds a new generalization mode `YICES_GEN_BY_PROJ_WIDE` to `yices_generalize_model{,_array}`. The historical pipeline extracts one literal implicant of `F(X, y)` at the model and projects it; the new mode instead

1. builds a polarity-aware Boolean abstraction of `F` against the model (model-false subterms collapsed to `FALSE`, AND/OR simplified in-place),
2. Tseitin-clausifies that abstraction into the y2sat solver,
3. SAT-guided-enumerates positive implicants, projects each through the unchanged legacy `get_implicant + project_literals` pipeline, and disjoins the results.

Two new exports ([yices_generalize_model_with_budget](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13159:0-13164:1) / `_array_with_budget`) expose a SAT-iteration cap; the existing exports keep their signatures and behave as `cube_budget = 0` (unbounded).

The wide path is **strictly opt-in**. `YICES_GEN_BY_PROJ` and `YICES_GEN_DEFAULT` retain their pre-PR semantics; only an explicit `YICES_GEN_BY_PROJ_WIDE` reaches the new code.

## 2. Review scope and approach

I read the full diff (`src/model/generalization.{c,h}`, the public-API and lock-free header changes, the API dispatch, the test file, and the PR description). I focused on:

- the soundness contract (`G(X)` true at the model, `G(X) ⇒ ∃y. F(X, y)`),
- the abstraction-builder lifetime and cache invalidation,
- the SAT enumeration loop (blocker discipline, termination, attempt counting),
- the failure taxonomy (assertable bugs vs. caller-precondition violations vs. per-cube projection errors),
- backward compatibility for every existing public-API caller,
- test coverage for the headline widening case.

I did not run the YicesQS / NRA-COMP benchmarks (out of scope per the PR text).

## 3. What this PR does well

- **Opt-in design.** After the latest rename, `YICES_GEN_BY_PROJ` and `YICES_GEN_DEFAULT` retain their pre-PR semantics byte-for-byte. The wide algorithm is only reachable through `YICES_GEN_BY_PROJ_WIDE`. This is the right rollout posture: zero risk to existing callers, free for them to opt in later.
- **Per-cube reuse of the legacy pipeline.** No new theory work. Each cube goes through `get_implicant + project_literals` unchanged, so the wide path inherits the legacy correctness for arithmetic projection. The novelty is the *enumeration of cubes*, not the projection of each one. This drastically narrows the soundness review surface.
- **Polarity-aware abstraction with model-false pruning.** A subtle and correct choice. Building the abstraction against the model up front (rather than the whole formula) means the SAT solver only ever explores model-true substructure, so the enumeration loop never wastes iterations on Boolean models that don't witness `F`. The `FALSE`-propagation through AND/OR keeps the abstraction small.
- **Attempt-counting budget, not success-counting.** A pathological `F` whose every implicant uses an unsupported literal would otherwise drag the loop through all `2^N` Boolean models before bailing. The attempt-counter (`num_attempts++`) bounds worst case directly; the PR description correctly justifies the choice (`generalization.c:1132`).
- **Failure taxonomy is well-considered.** Three categories — engine bugs (`assert`), caller-precondition violations (delegate to local for the specific error code), per-cube projection errors (skip-block-continue) — are documented inline at `generalization.c:1063-1090, 1136-1158` and in the PR. The "skip a cube but keep going" behavior is sound because every emitted cube was independently true at the model.
- **Blocker discipline.** The interaction with `nsat_solver_simplify_and_add_clause` is correctly guarded by `nsat_solver_prepare_for_next_search` (`generalization.c:1154-1158`). The inline comment explains why: y2sat's clause simplifier evaluates against the live model, which would empty the blocker outright. This is a subtle hazard that the author clearly thought through.
- **Hard regression on the headline case.** [test_overlapping_arith_disjunction](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:298:0-368:1) is now an `assert(0)` if wide collapses to local on `(y>x ∧ x>0) ∨ (z>x ∧ x>5)`. The feature can't silently regress: if widening stops working, the test suite fails loudly.
- **Documented output shape per mode.** The public-header docstring (`yices.h:4813-4829`) explicitly tells callers that `YICES_GEN_BY_PROJ` and `YICES_GEN_DEFAULT` keep their literal-per-element shape, and that `YICES_GEN_BY_PROJ_WIDE` may return a single `(or …)` term in `v->data[0]` when the enumerator emits multiple cubes. The new shape is a documented property of the new mode, not an unannounced change to existing modes.
- **Documentation.** The public-header docstring distinguishes the four modes; the internal [generalization.h](cci:7://file:///Users/sgl/git/yices/yices2/src/model/generalization.h:0:0-0:0) and [generalization.c](cci:7://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:0:0-0:0) headers describe both variants with the right "at least as broad, often strictly broader" phrasing (this PR went through a doc-overclaim pass and converged).

## 4. Findings

I did **not** find any soundness issues in the wide path.

### Minor

**4.1 — Pre-existing return-type oddity in [yices_generalize_model_array](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13212:0-13219:1) (not introduced here).**
`src/api/yices_api.c:13216` declares the wrapper as `EXPORTED term_t yices_generalize_model_array(...)`, but the public header at `src/include/yices.h:4844` declares it as `int32_t`. Since `term_t` is itself a `typedef int32_t` this compiles and is ABI-compatible, but the inconsistency is real. Inherited from commit `c222b4df`, not introduced by this PR. The new `_with_budget` wrapper at line 13222 correctly returns `int32_t`. **Not a blocker for this PR**, but worth a one-line fix in a follow-up.

**4.2 — `extra_error` not explicitly reset on entry.**
[gen_model_by_proj_sat_guided](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:1016:0-1229:1) passes `extra_error` straight through to nested calls and to [gen_model_by_proj_local](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:222:0-263:1) in the precondition-violation fallback (`generalization.c:1072, 1089, 1174, 1210`). If the caller passes uninitialized memory and the code path returns `0` without touching `*extra_error`, the caller sees stale data. The legacy [gen_model_by_proj_local](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:222:0-263:1) has the same property, and the public header documents `extra_error` as only meaningful when the call returns `GEN_PROJ_ERROR_UNSUPPORTED_ARITH_TERM`, so this is consistent with pre-existing behavior — but worth re-confirming in a future cleanup that all error-returning paths set `*extra_error`.

### Suggestions (non-blocking)

**4.3 — `XOR_TERM` / `EQ_TERM` as opaque leaves.**
The PR description acknowledges this is intentional; iff/xor decomposition would unlock widening on iff-style formulas. This is the obvious follow-up and the PR is right not to bundle it. A future PR should treat `XOR_TERM (a, b)` as `(a ∧ ¬b) ∨ (¬a ∧ b)` in the abstraction.

**4.4 — No microbenchmarks in-tree.**
The motivation is YicesQS performance, but the wide-vs-local cost trade-off isn't measured anywhere in the tree. The PR description defers this to "to be produced from the YicesQS side once this PR is merged." Acceptable, but a one-shot benchmark in `tests/` would make the cost story self-documenting for future maintainers. Not a blocker.

**4.5 — 1089-line single-file addition is dense.**
[src/model/generalization.c](cci:7://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:0:0-0:0) grows by ~24x. The new code is well-factored into named helpers ([abstract_signed](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:696:0-764:1), [clausify_node](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:805:0-856:1), [extract_implicant](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:905:0-936:1), [project_one_cube_into](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:267:0-303:1), [append_projected_cube_term](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:974:0-1014:1), [gen_model_by_proj_sat_guided](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:1016:0-1229:1)) and the inline comments are good. Still, future maintainers may benefit if `bool_dag_t` + `abs_builder_t` + the SAT-guided orchestrator move to a sibling file like `generalization_wide.c`. Not necessary for this PR; raise as a refactor.

### Open questions for the author

**4.6 — Cache invalidation on `ABS_ERROR`.**
The comment at `src/model/generalization.c:761-763` says:

> Roll back only the DAG; successful subcalls may have populated caches. The caller must discard this builder after `ABS_ERROR`.

The caller ([gen_model_by_proj_sat_guided](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:1016:0-1229:1)) does in fact discard the builder via [delete_abs_builder](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:462:0-468:1) immediately after `ABS_ERROR` and never re-uses it (the path falls into the local-fallback). This is correct as written. The question for the author is whether the comment should be promoted from "the caller must" to a stronger structural guarantee (e.g. an `is_poisoned` flag that asserts on any further use), so a future caller who reuses the builder after a partial error doesn't silently consume stale cache entries. Not a bug today; a future-proofing question.

**4.7 — Termination claim "Boolean enumeration is always finite".**
The blocker clause asserts the negation of the implicant just emitted. This rules out at least the Boolean assignment that produced it, so termination in `≤ 2^(atoms in abstraction)` iterations is correct. With per-cube `skip-and-continue`, a failed projection still adds the *implicant*'s blocker (`generalization.c:1158`, after the failure handler at 1136-1142). I confirmed by reading the loop that the blocker is added in all non-error exit paths, including projection-error skip. Worth double-checking with one more pair of eyes that no future code path adds a continue/break before the blocker.

## 5. Specific technical observations

### Abstraction ([abstract_signed](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:696:0-764:1), `bool_dag_t`)

- Sound: the model is consulted to prune false subterms; the resulting DAG is **propositionally equivalent to the model-true projection of F**. The SAT solver enumerates Boolean models of this projection, which are exactly the model-true Boolean implicants of `F`.
- Shared signed-atom handling via `bvar_to_atom` is a nice space optimization and correctly preserves polarity.
- `ABS_ERROR` precondition handling: if [eval_boolean_at_model](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:686:0-694:1) cannot evaluate a Boolean subterm, the abstraction aborts and the wide path delegates to [gen_model_by_proj_local](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:222:0-263:1). Per the PR text, "local is expected to fail the same way" — i.e. the fallback is purely to surface the canonical `GEN_EVAL_*` error code. Correct.

### `! builder.decomposed` short-circuit

The `decomposed` flag (`generalization.c:365, 736, 747`) is set on encountering `OR_TERM` **or** Boolean `ITE_TERM` *anywhere during the recursive walk*. When the entire DAG is OR-free and Boolean-ITE-free, the wide path short-circuits to [gen_model_by_proj_local](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:222:0-263:1) and returns the byte-identical legacy output. This is a stricter syntactic shortcut than "no top-level OR" — and that's the right one, because an OR buried under ANDs can still produce multi-cube output once the SAT loop runs.

### Clausification ([clausify_node](cci:1://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:805:0-856:1))

Standard Tseitin over an AND/OR/literal DAG; no concerns.

### SAT-guided loop

- Invariants are clear: `num_attempts` counts every iteration; `have_first` records whether any cube has been successfully projected.
- The three exit categories (`SAT exhausted | have_first`, `budget exhausted | have_first`, `no cube ever projected`) all dispatch to sound continuations. Verified by reading `generalization.c:1163-1218`.

### API dispatch ([yices_api.c](cci:7://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:0:0-0:0))

- New wrappers correctly `MT_PROTECT`-wrap the lock-free `_o_*_with_budget` implementations.
- The original `yices_generalize_model{,_array}` exports are reduced to thin wrappers that pass `cube_budget = 0`. This preserves ABI and avoids code duplication.
- One stylistic note: the lock-free [_o_yices_generalize_model_array](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13228:0-13231:1) is preserved as a thin forwarder. If nothing internal uses it any more, it could go in a follow-up; not a blocker.

### Tests

[tests/api/test_generalize_model.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:0:0-0:0) adds 9 cases (716 lines). Coverage breakdown:

| Test | Property |
|---|---|
| [test_pure_conjunction](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:264:0-295:1) | byte-identity wide ≡ local on conjunctive input |
| [test_overlapping_arith_disjunction](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:298:0-368:1) | **hard assert** wide strictly broader (headline case) |
| [test_nonoverlapping_disjunction](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:345:0-381:1) | soundness when only one disjunct is true |
| [test_array_form](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:410:0-497:1) | `_array` entry point |
| [test_sat_guided_compression](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:473:0-509:1) | SAT compression beats syntactic Cartesian |
| [test_sat_guided_no_sharing_parity](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:511:0-544:1) | no false compression on disjoint atoms |
| [test_sat_guided_budget_graceful](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:546:0-594:1) | `_with_budget` budget-exhaustion fallback |
| [test_shared_elimination_variable](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:622:0-662:1) | joint Cartesian projection |
| [test_sat_guided_polarity](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_generalize_model.c:664:0-696:1) | signed-literal + model-false pruning |

Every test verifies the contract `each cube true at the model` and `AND(local) ⇒ AND(wide)`. The headline widening case is a hard assert, not a warning — this is exactly the right invariant to lock in. I'd suggest adding **one** more test in a follow-up: an iff-style input where wide currently *cannot* widen (since `EQ_TERM` is opaque), to pin down the current limitation as a documented "TODO" via a soft-warn assertion.

## 6. Documentation

- `src/include/yices.h:4763-4870`: clear mode table; the "strictly broader on many such inputs, but not on every such input" phrasing is well-calibrated. The output-shape paragraph (`4813-4829`) spells out exactly which modes guarantee the literal-per-element shape and what `YICES_GEN_BY_PROJ_WIDE` may return.
- `src/include/yices_types.h:282-304`: enum docstring matches.
- [src/model/generalization.h](cci:7://file:///Users/sgl/git/yices/yices2/src/model/generalization.h:0:0-0:0) and [src/model/generalization.c](cci:7://file:///Users/sgl/git/yices/yices2/src/model/generalization.c:0:0-0:0) file-top headers: consistent after the recent doc-pass cleanup. No remaining "wide is the public default" drift.
- The PR description (now ~155 lines) is appropriately concise, has a correct mode table, documents the new `_with_budget` API, and lists out-of-scope items.

## 7. Backward compatibility

- **Source-level**: ✓ — no existing signature changes.
- **ABI**: ✓ — `YICES_GEN_BY_PROJ_WIDE` is appended to the enum after the existing values, so the numeric encoding of `YICES_GEN_DEFAULT`, `YICES_GEN_BY_SUBST`, `YICES_GEN_BY_PROJ` is unchanged.
- **Behavior**: ✓ — `YICES_GEN_BY_PROJ`, `YICES_GEN_BY_SUBST`, and `YICES_GEN_DEFAULT` retain pre-PR semantics, *including* the documented literal-per-element shape of `v`. Confirmed by reading the API dispatch and by running the existing API test suite. The new output-shape branch (a single `(or …)` term in `v->data[0]`) is only reachable when the caller explicitly selects `YICES_GEN_BY_PROJ_WIDE` and is documented in that mode's own docstring.
- **`efsolver`**: untouched. It has its own implicant-then-project pipeline.

## 8. Performance considerations

- Per-cube projection is the same legacy pipeline at the same per-cube cost; the wide overhead is one DAG construction + one Tseitin pass + one SAT loop per call. For OR-free / Boolean-ITE-free input the wide path short-circuits, so callers that pass such `F` under `_WIDE` pay essentially zero extra.
- The SAT solver instance is local to each call (`sat_solver_t sat` on the stack at `generalization.c:1023`). No global solver-state pollution.
- Budget: `cube_budget = 0` (unbounded) is the default for the legacy exports. Callers concerned about pathological inputs should use `_with_budget` with a small cap (the test uses 4). No benchmark numbers yet, as noted.

## 9. Recommendation

**Approve for merge**, subject to the author's judgement on the minor items above. The PR is:

- correct on the soundness contract,
- strictly opt-in (no risk to existing callers),
- well tested (with the headline widening case locked in as a hard assert),
- well documented (after the recent docs cleanup),
- reasonably scoped (single feature, no drive-by changes).

The follow-ups I'd track in a separate issue:

1. iff/xor decomposition in the abstraction (4.3),
2. an `is_poisoned` flag on `abs_builder_t` to make the post-`ABS_ERROR` invariant structural (4.6),
3. one extra test pinning the current iff limitation (§5),
4. fix the pre-existing `term_t` return-type oddity in [yices_generalize_model_array](cci:1://file:///Users/sgl/git/yices/yices2/src/api/yices_api.c:13212:0-13219:1) (4.1),
5. a microbenchmark once YicesQS has data (4.4),
6. consider splitting the wide implementation into a sibling file (4.5).

None of these are blockers.

disteph:
## Evaluation of Wide Projections on NRA Benchmarks

I evaluated the impact of the new wide-projection option on the SMT-LIB NRA benchmark set. The goal was to measure the cases where Yices actually makes projection calls, rather than averaging over the full benchmark set where most instances do not exercise this code path. Evaluation has been supported by Codex/GPT5.5.

### Instrumentation and Experiment Branches

The experiment used three auxiliary branches:

- `SRI-CSL/yices2`, branch `wide-projection-stats`
  - Adds env-gated instrumentation in `src/model/generalization.c`.
  - When `YICES_WIDE_PROJECTION_STATS=1` is set, Yices prints one stats line per wide-projection call.

- `SRI-CSL/yices2_ocaml_bindings`, branch `wide-projection-stats`
  - Points `vendor/yices2` at the instrumented Yices commit.

- `disteph/yicesQS`, branch `wide-projection-stats`
  - Adds scripts to run NRA benchmarks, collect projection-call statistics, and compare regular projection against `-wide-projection 0`.

These branches are for experiment reproducibility only; they are not intended to be merged into the main branches.

### Projection-Call Coverage

Using the Yices instrumentation, I first identified the NRA instances that make at least one projection call.

- Total NRA benchmarks considered: 3819
- Benchmarks making at least one projection call: 424
- Total projection calls on those benchmarks: 1648
- Total projected cubes: 5548
- Average cubes per projection call: 3.37
- No fallback calls were observed in this run

The cube distribution is skewed: many calls project only one or two cubes, but a meaningful subset projects substantially more. The instances with the highest cubes-per-call ratios are mostly Keymaera benchmarks, especially `magnetic_field-*` and related proof-node families.

### Runtime Comparison

I compared regular projection against `wide-projection 0` on exactly the 424 benchmarks that made projection calls.

Two runtime datasets were considered:

1. A local 5-second timeout run using the experiment harness in `yicesQS`.
2. A separate 5-minute timeout experiment from `NRA5.xlsx`, comparing the same two configurations.

The 5-second run showed a very strong advantage for wide projections:

| metric | value |
|---|---:|
| projection-call instances | 424 |
| completed by both configurations | 417 |
| timed out by both configurations | 7 |
| wide faster among completed | 416 |
| wide slower among completed | 1 |
| regular total runtime | 22.421s |
| wide-projection 0 total runtime | 10.752s |
| total runtime reduction | 52.05% |

The independent 5-minute run gives a more conservative but still positive result:

| metric | value |
|---|---:|
| matched projection-call instances | 424 |
| solved by both configurations | 418 |
| timed out by both configurations | 6 |
| wide faster among solved | 272 |
| regular faster among solved | 146 |
| regular total runtime | 7.776s |
| wide-projection 0 total runtime | 6.442s |
| total runtime reduction | 17.15% |

There were no result disagreements in the matched data: solved instances had matching `sat`/`unsat` results, and timeout cases timed out in both configurations.

### Dependence on Cubes per Call

The strongest evidence for wide projections appears exactly where expected: instances with many cubes per projection call.

For the 5-minute experiment, grouped by average cubes per projection call:

| avg cubes/call | instances | regular time | wide0 time | runtime change | wide faster / slower |
|---|---:|---:|---:|---:|---:|
| `>= 9` | 15 | 0.603s | 0.183s | -69.56% | 15 / 0 |
| `7.5 - 8.99` | 38 | 0.633s | 0.476s | -24.71% | 22 / 16 |
| `5 - 7.49` | 168 | 2.310s | 2.013s | -12.86% | 105 / 63 |
| `3 - 4.99` | 70 | 1.190s | 0.893s | -24.97% | 54 / 16 |
| `2 - 2.99` | 45 | 0.841s | 0.731s | -13.02% | 31 / 14 |
| `< 2` | 82 | 2.200s | 2.146s | -2.49% | 45 / 37 |

The high-cube bucket is particularly clear: all 15 instances with average cubes/call at least 9 were faster with wide projections, with a 69.6% aggregate runtime reduction.

### How to Rerun the Experiment

The setup assumes this directory layout:

```text
yices/
  yicesQS/
  yices2_ocaml_bindings/
  SMTLib/NRA/
```

Check out the experiment branches:

```sh
cd /path/to/yices/yices2_ocaml_bindings
git checkout wide-projection-stats
git submodule update --init --recursive

cd vendor/yices2
git checkout wide-projection-stats

cd /path/to/yices/yicesQS
git checkout wide-projection-stats
```

Build the instrumented vendored Yices and yicesQS:

```sh
cd /path/to/yices/yices2_ocaml_bindings
YICES2_FORCE_LOCAL=1 dune build @vendor_deps

cd /path/to/yices/yicesQS
make build
```

Collect projection-call statistics on NRA with `wide-projection 0`:

```sh
cd /path/to/yices/yicesQS

NRA_WIDE_STATS_BUDGETS=0 \
NRA_WIDE_STATS_TIMEOUT=5 \
NRA_WIDE_STATS_OUT_DIR=experiments/nra-wide-stats/nra-wide0 \
make nra-wide-stats
```

This produces:

```text
experiments/nra-wide-stats/nra-wide0/runs.tsv
experiments/nra-wide-stats/nra-wide0/projection_calls.tsv
experiments/nra-wide-stats/nra-wide0/summary.tsv
```

Extract the instances that made projection calls, sorted by decreasing average cubes per call:

```sh
TAB=$(printf '\011')

awk -F "$TAB" '
  BEGIN {
    OFS = "\t"
  }
  NR == 1 {
    next
  }
  $2 == "0" {
    calls[$1] += 1
    sum[$1] += $10
    if ($10 > max[$1]) {
      max[$1] = $10
    }
  }
  END {
    for (f in calls) {
      print f, sprintf("%.6f", sum[f] / calls[f]), max[f], calls[f]
    }
  }
' experiments/nra-wide-stats/nra-wide0/projection_calls.tsv \
| sort -t "$TAB" -k2,2nr -k1,1 \
> experiments/nra-wide-stats/nra-wide0/projection_instances.tsv
```

Compare regular projection against `wide-projection 0` on just those instances:

```sh
scripts/compare_nra_projection_runtimes.sh \
  --no-build \
  --instances experiments/nra-wide-stats/nra-wide0/projection_instances.tsv \
  --timeout 5 \
  --out-dir experiments/nra-wide-runtime/nra-wide0-vs-regular
```

For a 5-minute timeout run, use `--timeout 300`.

The runtime comparison output is:

```text
experiments/nra-wide-runtime/nra-wide0-vs-regular/runs.tsv
experiments/nra-wide-runtime/nra-wide0-vs-regular/comparison.tsv
```

In `comparison.tsv`, `delta_s = regular_s - wide0_s`, so positive values mean wide projection was faster. `pct_change` is computed as `100 * (wide0_s - regular_s) / regular_s`, so negative values mean runtime decreased.

### Conclusion

These results support keeping wide projections in Yices.

The benefit is not uniform on every individual benchmark. Some very small instances become slightly slower, and many of the per-instance differences are in the millisecond range. However, the aggregate result is positive, and the improvement is concentrated in precisely the cases where the feature should help: projection calls that produce multiple cubes.

This suggests that wide projection is a useful capability to have in Yices, especially as an optional or tunable projection mode. It provides a clear win on the projection-heavy NRA cases without changing solver results in the evaluated data.

PR #631, opened 2026-05-26
Fix y2sat delegate multi-check

disteph:
## Summary

Fixes y2sat delegate append mode when a context performs multiple SAT checks and appends new clauses after a previous `SAT` result.

Previously, y2sat kept the previous complete SAT assignment and `STAT_SAT` status. If the QF_BV delegate path appended new clauses afterward, those clauses could be simplified against the stale assignment. In the bad case, a newly added clause that was false under the old model could be reduced to an empty clause, causing the next check to return `UNSAT` even though the accumulated formula was still satisfiable.

## Fix

Adds `nsat_solver_prepare_for_next_search()` to y2sat. It prepares the solver for post-SAT clause addition by:

- backtracking to decision level 0;
- clearing transient conflict/search state;
- resetting the decision-variable list;
- setting status back to `STAT_UNKNOWN` unless an empty clause is already present;
- preserving problem clauses and learned clauses.

The y2sat delegate wrappers now call this preparation hook before adding clauses when the solver is currently in `STAT_SAT`.

## Tests

Adds regression coverage in `tests/api/test_context_delegate_stats.c`:

- a QF_BV append-mode multi-check using `sat-delegate=y2sat`;
- a focused y2sat delegate multi-check that appends a clause after a SAT result and checks again.

The focused regression fails without the delegate prepare hook: the second check incorrectly returns `UNSAT`. With this fix, it returns `SAT`.

## Verification

Ran:

```sh
make MODE=release check-api
```

Result:

```text
Total: 35
Pass:  35
Skip:  0
Fail:  0
```

Supported by Codex/GPT5.5

### DIFF
--- a/src/solvers/cdcl/delegate.c
+++ b/src/solvers/cdcl/delegate.c
@@ -42,17 +42,26 @@
 /*
  * WRAPPERS FOR THE YICES SAT_SOLVER
  */
+static void ysat_prepare_to_add_clause(sat_solver_t *solver) {
+  if (solver->status == STAT_SAT) {
+    nsat_solver_prepare_for_next_search(solver);
+  }
+}
+
 static void ysat_add_empty_clause(void *solver) {
+  ysat_prepare_to_add_clause(solver);
   nsat_solver_simplify_and_add_clause(solver, 0, NULL);
 }
 
 static void ysat_add_unit_clause(void *solver, literal_t l) {
+  ysat_prepare_to_add_clause(solver);
   nsat_solver_simplify_and_add_clause(solver, 1, &l);
 }
 
 static void ysat_add_binary_clause(void *solver, literal_t l1, literal_t l2) {
   literal_t a[2];
 
+  ysat_prepare_to_add_clause(solver);
   a[0] = l1;
   a[1] = l2;
   nsat_solver_simplify_and_add_clause(solver, 2, a);
@@ -61,6 +70,7 @@ static void ysat_add_binary_clause(void *solver, literal_t l1, literal_t l2) {
 static void ysat_add_ternary_clause(void *solver, literal_t l1, literal_t l2, literal_t l3) {
   literal_t a[3];
 
+  ysat_prepare_to_add_clause(solver);
   a[0] = l1;
   a[1] = l2;
   a[2] = l3;
@@ -68,6 +78,7 @@ static void ysat_add_ternary_clause(void *solver, literal_t l1, literal_t l2, li
 }
 
 static void ysat_add_clause(void *solver, uint32_t n, literal_t *a) {
+  ysat_prepare_to_add_clause(solver);
   nsat_solver_simplify_and_add_clause(solver, n, a);
 }
 
--- a/src/solvers/cdcl/delegate.c
+++ b/src/solvers/cdcl/delegate.c
@@ -44,6 +44,12 @@
  */
 static void ysat_prepare_to_add_clause(sat_solver_t *solver) {
   if (solver->status == STAT_SAT) {
+    /*
+     * y2sat append-after-SAT is supported only for the incremental delegate,
+     * which disables preprocessing. External incremental delegates handle
+     * post-SAT clause additions internally.
+     */
+    assert(! solver->preprocess);
     nsat_solver_prepare_for_next_search(solver);
   }
 }
--- a/src/solvers/cdcl/delegate.c
+++ b/src/solvers/cdcl/delegate.c
@@ -42,14 +42,30 @@
 /*
  * WRAPPERS FOR THE YICES SAT_SOLVER
  */
+#ifndef NDEBUG
+static bool ysat_has_no_preprocessing_eliminations(const sat_solver_t *solver) {
+  uint32_t i, n;
+
+  n = solver->nvars;
+  for (i=0; i<n; i++) {
+    if (solver->ante_tag[i] == ATAG_PURE || solver->ante_tag[i] == ATAG_ELIM) {
+      return false;
+    }
+  }
+  return true;
+}
+#endif
+
 static void ysat_prepare_to_add_clause(sat_solver_t *solver) {
   if (solver->status == STAT_SAT) {
     /*
-     * y2sat append-after-SAT is supported only for the incremental delegate,
-     * which disables preprocessing. External incremental delegates handle
-     * post-SAT clause additions internally.
+     * nsat_solver_simplify_and_add_clause rewrites ATAG_SUBST literals.
+     * We rely on incremental y2sat running without preprocessing, so
+     * resolution-eliminated variables (ATAG_ELIM) and pure variables
+     * (ATAG_PURE) cannot appear in later appended clauses.
      */
     assert(! solver->preprocess);
+    assert(ysat_has_no_preprocessing_eliminations(solver));
     nsat_solver_prepare_for_next_search(solver);
   }
 }
--- a/src/solvers/cdcl/new_sat_solver.c
+++ b/src/solvers/cdcl/new_sat_solver.c
@@ -8773,6 +8773,22 @@ static void backtrack(sat_solver_t *solver, uint32_t back_level) {
 }
 
 
+/*
+ * Prepare for adding more problem clauses after a SAT result.
+ */
+void nsat_solver_prepare_for_next_search(sat_solver_t *solver) {
+  if (solver->decision_level > 0) {
+    backtrack(solver, 0);
+  }
+  reset_var_list(&solver->list);
+  solver->backtrack_level = 0;
+  solver->conflict_tag = CTAG_NONE;
+  if (! solver->has_empty_clause) {
+    solver->status = STAT_UNKNOWN;
+  }
+}
+
+
 #if 0
 
 // DISABLED FOR TESTING
--- a/src/solvers/cdcl/new_sat_solver.c
+++ b/src/solvers/cdcl/new_sat_solver.c
@@ -2273,15 +2273,6 @@ static void extend_var_list(nvar_list_t *list, uint32_t n) {
 
 
 /*
- * Increase the number of variables to n
- */
-static void var_list_add_vars(nvar_list_t *list, uint32_t n) {
-  assert(list->nvars <= n && n <= list->size);
-  list->nvars = n;
-}
-
-
-/*
  * Delete
  */
 static void delete_var_list(nvar_list_t *list) {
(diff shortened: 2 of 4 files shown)


disteph:
# Review by Windsurf/Opus4.7: Fix y2sat delegate multi-check

## Summary

The PR fixes a real soundness bug in the y2sat delegate's append/multi-check mode: clauses appended after a SAT result were simplified against the previous (still-installed, decision-level > 0) assignment, occasionally turning a satisfiable formula into a fake empty-clause UNSAT. The change is well-scoped, addresses the root cause upstream in the SAT solver instead of layering a workaround in the context layer, and is exercised by four regression tests. LGTM.

## Root cause

[nsat_solver_simplify_and_add_clause](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3788:0-3869:1) simplifies each appended literal against `lit_value(solver, l)`:

```c
case VAL_FALSE:
  break;
case VAL_UNDEF_FALSE:
case VAL_UNDEF_TRUE:
  lit[j++] = l;
  break;
default: // true literal -> clause is true, drop it
  return;
```

When `solver->status == STAT_SAT`, all decision-level > 0 assignments are still in `solver->value`. A newly added clause whose literals are all `VAL_FALSE` under the previous model collapses to `n == 0` and adds the empty clause; the next [nsat_solve](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:11202:0-11318:1) short-circuits with `STAT_UNSAT`.

A related second hazard exists for variables tagged `ATAG_SUBST` by [try_scc_simplification](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:9831:0-9913:1) (which runs even with `preprocess=false`). After SCC substitution, the substituted variable is removed from clauses, then its model value is later set by [extend_assignment](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:9971:0-9987:1) after SAT. Appended clauses mentioning that variable would either be wrongly simplified against the extended-model value or, even if added, never propagate to the survivor — leading to an incorrect SAT model after the next check.

## Fix shape

Two cooperating changes, both in the SAT solver layer:

1. **[nsat_solver_prepare_for_next_search](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:8822:0-8835:1)** — called from each `ysat_add_*_clause` wrapper when `solver->status == STAT_SAT`. Performs the minimum necessary to make [nsat_solver_simplify_and_add_clause](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3788:0-3869:1) and a subsequent [nsat_solve](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:11202:0-11318:1) safe and idempotent: backtrack to level 0, clear the conflict tag, clear the transient `active_vars` scratch buffer, and reset `status` to `STAT_UNKNOWN` unless an empty clause is already present. Importantly, the decision-variable list and per-variable rank are preserved, so VSIDS-style ordering carried from the previous search survives.

2. **`ATAG_SUBST` rewriting at the top of [nsat_solver_simplify_and_add_clause](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3788:0-3869:1)** — every incoming literal is passed through [full_lit_subst](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3773:0-3785:1), which follows the substitution chain to the survivor. The dedup/opposite/false-removal pass then operates on canonical literals. The function is a no-op for the first solve (no substitutions exist yet) and otherwise corrects any literal over a now-eliminated variable.

In addition, [var_list_add_vars](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:2407:0-2428:1) is extended so that variables added between solves are appended to the existing list (with `unassigned_index` advanced to the tail, so fresh variables are tried first), and [nsat_solve](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:11202:0-11318:1)/[nsat_apply_preprocessing](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:11146:0-11199:1) only populate the list if it is empty.

## Correctness

I walked all the entry paths and the state machine is sound:

- **First solve.** Variable list is empty; [var_list_add_all_if_empty](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:2464:0-2472:1) populates it as before. [full_lit_subst](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3773:0-3785:1) is the identity (no `ATAG_SUBST` tags exist yet). Behaviour is unchanged from the previous code.
- **SAT → append → recheck.** The wrapper detects `STAT_SAT`, calls [nsat_solver_prepare_for_next_search](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:8822:0-8835:1). [backtrack(0)](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:8781:0-8819:1) undoes all level > 0 assignments (the existing routine also updates `unassigned_index` to the highest-ranked unassigned var, so the search heap state is consistent). `status` becomes `STAT_UNKNOWN`. The new clause is rewritten through [full_lit_subst](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3773:0-3785:1), simplified against only the level-0 assignments (which are unconditionally implied), and added. The next [nsat_solve](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:11202:0-11318:1) skips the list-populate (non-empty list), runs simplification, and searches with the previously-learned variable order plus any tail-priority new variables.
- **SAT → add_var → append → recheck.** Variable creation is intentionally not wrapped, but appending new vars while `status == STAT_SAT` is harmless: new vars get `VAL_UNDEF_*` initial value, are inserted at the tail of the list, and the prior level > 0 stale assignments are cleaned up by the prepare hook when the next clause arrives.
- **SUBST case.** A new clause containing a substituted variable is rewritten to refer to the survivor before simplification. The survivor's `value` slot was unassigned by [backtrack(0)](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:8781:0-8819:1) (it was a regular decision/propagation, not a forced post-SAT assignment), so simplification keeps the literal. After the next SAT result, [extend_assignment](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:9971:0-9987:1) re-derives the substituted variable's value from the survivor.

The `ATAG_SUBST` rewriting is also safe at every other call site of [nsat_solver_simplify_and_add_clause](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3788:0-3869:1) (e.g. [yices_sat_new.c](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/yices_sat_new.c:0:0-0:0) DIMACS loading): the solver is fresh, no substitutions exist, and the pass is O(n) ante_tag lookups per clause.

Defensive debug-only check [ysat_has_no_preprocessing_eliminations](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/delegate.c:45:0-55:1) rules out `ATAG_PURE`/`ATAG_ELIM` tags in the wrapper, which is the precise invariant [full_lit_subst](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3773:0-3785:1) relies on (it only handles `ATAG_SUBST`). Combined with the runtime guarantee that the incremental delegate always runs with `preprocess=false`, this is the right contract.

## Tests

Four new tests cover:

1. **`check_y2sat_append_add_sat_clause_after_solve_case`** — end-to-end QF_BV through the context layer, asserts `append_checks == 2 && rebuild_checks == 0` after a single post-SAT append.
2. **`check_y2sat_append_three_round_blocking_case`** — three successive blocking-clause rounds via the context layer, asserts `append_checks == 4 && rebuild_checks == 0`. Good coverage for state leaking across more than one round trip.
3. **`check_y2sat_delegate_multi_check_add_clause_after_sat_case`** — focused white-box test through the raw delegate API, constructs the blocking clause from the actual model values so it stresses the `STAT_SAT → simplify-against-stale-model` path regardless of which model the first check picks.
4. **`check_y2sat_delegate_append_substituted_literal_case`** — exercises the `ATAG_SUBST` path by asserting `x1 ↔ x2` (two binary clauses), then appending a unit clause that negates the substituted variable's model value. Sensibly degrades to a no-op early return if the SCC heuristic stops producing a substitution on this input.

## Suggestions (non-blocking)

- **Docstring for [var_list_add_vars](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:2407:0-2428:1)**: mention that `unassigned_index` is moved to the tail when the list is already populated, so future maintainers understand the intentional "new variables decided first" heuristic.
- **`block_current_model_on_atoms` helper**: hard-coded `n <= 4` is fine for the current tests; consider a brief comment so the magic constant doesn't get lifted into new tests by accident.
- **Test layering**: the two delegate-direct tests and the two context-level tests cover different layers; a one-line comment in each indicating which layer it exercises would help triage when something regresses.
- **[nsat_solver_simplify_and_add_clause](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3788:0-3869:1) performance**: the new [full_lit_subst](cci:1://file:///Users/sgl/git/yices/yices2/src/solvers/cdcl/new_sat_solver.c:3773:0-3785:1) pass is one extra ante_tag read per literal on every clause add. Negligible for SMT workloads, but worth keeping in mind for any future bulk-load fast path in [yices_sat_new.c](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/yices_sat_new.c:0:0-0:0).

## Verdict

Approve. The fix targets the right place, is minimal, fully documented in the header comments, defensively asserted in debug builds, and well-regressed. The decision to keep the variable list across multi-checks (rather than reset it) preserves search heuristics and is a strict improvement over the simpler "reset everything" approach. The combination of backtrack-to-level-0 plus `ATAG_SUBST` literal rewriting cleanly closes both soundness holes.

PR #632, opened 2026-05-26
mcsat: tighten trail cache update and GC sweep loops

ahmed-irfan:
Two small perf cleanups in src/mcsat/trail.c, no behavior change:

  - **`trail_update_extra_cache`**: replace the two O(|cache|) loops with a single pass over `trail->elements` (O(|trail|)), fused so both caches
   share one `trail_get_value` call.
  - **`trail_gc_sweep`**: collapse the three per-cache loops into one — each unmarked variable now costs one `gc_info_get_reloc` lookup instead
  of three.

### DIFF
--- a/src/mcsat/trail.c
+++ b/src/mcsat/trail.c
@@ -306,30 +306,24 @@ void trail_gc_mark(mcsat_trail_t* trail, gc_info_t* gc_vars) {
 }
 
 void trail_gc_sweep(mcsat_trail_t* trail, const gc_info_t* gc_vars) {
-  variable_t var;
-
   assert(gc_vars->is_id);
 
-  // Remove from the model cache, otherwise the cache might contain wrongly
-  // typed variables
-  for (var = 0; var < trail->model.size; ++ var) {
-    if (var != variable_null && gc_info_get_reloc(gc_vars, var) == variable_null) {
-      assert(!trail_has_value(trail, var));
-      if (mcsat_model_has_value(&trail->model, var)) {
-        mcsat_model_unset_value(&trail->model, var);
-      }
-      assert(!trail_has_value(trail, var));
+  // model, target_cache, and best_cache grow in lockstep with trail->level
+  // via trail_new_variable_notify; iterate the common range and clean every
+  // unmarked variable from all three caches in one pass. Skip var 0 (the
+  // variable_null sentinel).
+  for (uint32_t var = 1; var < trail->model.size; ++var) {
+    if (gc_info_get_reloc(gc_vars, var) != variable_null) {
+      continue;
     }
-  }
-  for (var = 0; var < trail->target_cache.size; ++ var) {
-    if (var != variable_null && gc_info_get_reloc(gc_vars, var) == variable_null &&
-	mcsat_model_has_value(&trail->target_cache, var)) {
+    assert(!trail_has_value(trail, var));
+    if (mcsat_model_has_value(&trail->model, var)) {
+      mcsat_model_unset_value(&trail->model, var);
+    }
+    if (mcsat_model_has_value(&trail->target_cache, var)) {
       mcsat_model_unset_value(&trail->target_cache, var);
     }
-  }
-  for (var = 0; var < trail->best_cache.size; ++ var) {
-    if (var != variable_null && gc_info_get_reloc(gc_vars, var) == variable_null &&
-	mcsat_model_has_value(&trail->best_cache, var)) {
+    if (mcsat_model_has_value(&trail->best_cache, var)) {
       mcsat_model_unset_value(&trail->best_cache, var);
     }
   }
--- a/src/mcsat/trail.c
+++ b/src/mcsat/trail.c
@@ -427,23 +427,25 @@ void trail_recache(mcsat_trail_t* trail, uint32_t round) {
 
 void trail_update_extra_cache(mcsat_trail_t* trail) {
   // update target and best cache w.r.t. current trail if the trail size is bigger
-  variable_t var;
-  if (trail->elements.size > trail->best_depth) {
-    // keep only the values for assigned variables as best promising assignment
-    for (var = 0; var < trail->best_cache.size; ++var) {
-      if (trail_has_value(trail, var)) {
-        mcsat_model_set_value(&trail->best_cache, var, trail_get_value(trail, var));
-      }
+  const bool update_best   = trail->elements.size > trail->best_depth;
+  const bool update_target = trail->elements.size > trail->target_depth;
+  if (!update_best && !update_target) {
+    return;
+  }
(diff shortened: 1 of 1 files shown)


disteph:
For the record: Code Review by Windsurf/Opus4.7 

# PR #632 (mcsat: tighten trail cache update and GC sweep loops)

## Summary

Two well-motivated micro-optimizations in `@/Users/sgl/git/yices/yices2/src/mcsat/trail.c`. The changes are **correct** under existing invariants and are net wins. A few minor nits / suggestions below; nothing blocking.

## Correctness review

### [trail_update_extra_cache](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:427:0-448:1) (lines 428–449)

Old loops iterated `[0, *_cache.size)` and filtered with `trail_has_value(var)`. New code iterates `trail->elements`. Equivalence requires:

> `{var : trail_has_value(var)}` == `set(trail->elements)` at every call site.

`trail_has_value` is true iff `trail->index.data[var] != -1`. Looking at `@/Users/sgl/git/yices/yices2/src/mcsat/trail.c:257-273` ([trail_pop_propagation](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:256:0-272:1)), a popped lower-level propagation is removed from `elements` but its `index` is **not** reset — it sits in `to_repropagate` with `trail_has_value() == true`. So the equivalence fails iff `to_repropagate` is non-empty. The PR's commit message acknowledges this as a "transient".

Verified the three call sites in `@/Users/sgl/git/yices/yices2/src/mcsat/solver.c`:

- **`@/Users/sgl/git/yices/yices2/src/mcsat/solver.c:1470`** — called after the [trail_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:274:0-289:1) loop in [mcsat_backtrack_to](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:1448:0-1470:1). Each [trail_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:274:0-289:1) either drains `to_repropagate` via [trail_pop_decision](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:225:0-242:1) (`@/Users/sgl/git/yices/yices2/src/mcsat/trail.c:226-243`) or never populates it (the fake-push branch). So `to_repropagate.size == 0` here. ✓
- **`@/Users/sgl/git/yices/yices2/src/mcsat/solver.c:1526`** — restart path; same story. ✓
- **`@/Users/sgl/git/yices/yices2/src/mcsat/solver.c:2905`** — guarded by [trail_clear_cache](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:457:0-460:1) immediately before; trail state has no in-flight repropagations. ✓

So the change is safe **today**, but it now silently relies on this invariant. Suggest adding either a comment or an inexpensive guard at the top of the function:

    assert(trail->to_repropagate.size == 0);

That makes the new precondition explicit and costs nothing in release builds.

### [trail_gc_sweep](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:307:0-335:1) (lines 308–336)

The fusion is predicated on `model.size == target_cache.size == best_cache.size`. The only growth path is [trail_new_variable_notify](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:84:0-96:1) (`@/Users/sgl/git/yices/yices2/src/mcsat/trail.c:85-97`), which calls `mcsat_model_new_variable_notify` on all three identically. ✓

However, `mcsat_model_set_value` will *also* grow a model on demand (`@/Users/sgl/git/yices/yices2/src/mcsat/model.c:120-123`). All call sites I checked invoke `mcsat_model_set_value` only on variables previously passed through [trail_new_variable_notify](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:84:0-96:1), so the three sizes stay equal in practice. Recommend an assertion to make the invariant explicit:

    assert(trail->model.size == trail->target_cache.size);
    assert(trail->model.size == trail->best_cache.size);

Otherwise a future caller that grows one cache without the others would silently leak unmarked entries past the new loop bound.

Skipping `var == 0` (= `variable_null`, see `@/Users/sgl/git/yices/yices2/src/mcsat/variable_db.h:32-34`) is safe — no real variable is ever `0`. The dropped trailing `assert(!trail_has_value(trail, var))` is genuinely redundant with the one above the unset (it queries `trail->index`, not the model, so unsetting the model can't change it). ✓

## Style / minor nits

- **Loop-variable type change**: both new loops switched from `variable_t` (`int32_t`) to `uint32_t`. `gc_info_get_reloc` expects `variable_t`, so each call now does an implicit `uint32_t → int32_t` conversion. Functionally fine but inconsistent with the surrounding code (e.g. [trail_copy_unassigned_cache](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:373:0-381:1) at `@/Users/sgl/git/yices/yices2/src/mcsat/trail.c:374-382` uses `variable_t`). Prefer `variable_t var = 1;` for consistency, or `for (variable_t var = 1; var < (variable_t) trail->model.size; ++var)`.
- **Comment wording**:

      // model, target_cache, and best_cache grow in lockstep with trail->level
      // via trail_new_variable_notify; iterate the common range and clean every
      // unmarked variable from all three caches in one pass. Skip var 0 (the
      // variable_null sentinel).

  They grow in lockstep with the **variable database**, not with `trail->level` — those are unrelated.
- **`update_best`/`update_target` flags**: nice readability win, no objection.

## Performance

- [trail_update_extra_cache](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:427:0-448:1): clear win — `O(|trail|)` instead of `O(|max var index|)`, fused so the two writes share one `trail_get_value`. Both writes also avoid the per-iteration `trail_has_value` branch.
- [trail_gc_sweep](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:307:0-335:1): clear win — one `gc_info_get_reloc` per variable instead of three, and one pass through memory instead of three.

## Recommendation

**Approve** with optional follow-ups:

1. Add `assert(trail->to_repropagate.size == 0)` at the top of [trail_update_extra_cache](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:427:0-448:1).
2. Add the size-equality asserts in [trail_gc_sweep](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.c:307:0-335:1).
3. Restore `variable_t` for the loop variable.
4. Fix the "grow in lockstep with trail->level" comment.

None of the above are required for merging; they just future-proof the implicit invariants the new code relies on.

Issue #390, opened 2021-11-06
Requesting improved smt2 array support


GreenBeard:
I understand that yices doesn't differentiate between arrays, and functions internally, but I was wondering if the output while using the `--smt2-model-format` flag could have an option to output the result as an array if the input was an array (or as a function if the input was a function)? Currently, the output being a function when the input was an array makes it harder to use the result in some of the other solvers (e.g. z3, and cvc5) as they seem to differentiate between the two.

Current output:
```
(define-fun params ((x!0 Int)) Int (ite (= x!0 6) 5 (ite (= x!0 5) 2 (ite (= x!0 4) 2 (ite (= x!0 3) 0 (ite (= x!0 2) 1 14))))))
```

Desired output (I hand translated this, so I may have made minor mistakes, but I think it is correct):
```
(define-fun params () (Array Int Int) (store (store (store (store (store ((as const (Array Int Int)) 14) 2 1) 3 0) 4 2) 5 2) 6 5))
```

Ideally, yices would also support `(as const (Array Int Int))` inputs too, but I assume that would be more work. Unfortunately, this means that a separate flag probably is needed (to enable outputting arrays as arrays) as otherwise yices could produce smtlib2 formats that it couldn't read.

Edit: The `((as const (Array Int Int)) 14)` syntax appears to be non-standard smt-lib syntax that cvc4, z3, mathsat, and possibly more provers support. I would still appreciate this being supported, but it doesn't appear to be specified in the standards anywhere that I can find, and I see some comments elsewhere on the internet saying it isn't part of the standard. :/

disteph:
I suspect `(as const (Array Int Int))` would be some work, starting with an extension of the SMT2 parser. We may in the near future equip Yices with a generated parser instead of the currently hard-coded one, which would make it easy to do such extensions.

But I suspect the more involved work would be to tweak the algorithm for the array/uninterpreted function theory in order to set the default return value of a lambda-function to whatever is specified in the `(as const (Array Int Int))` constructs, and trigger backtracks when hitting conflictual information that essentially amount to `(= (as const (Array Int Int) 0) (as const (Array Int Int) 1)`.

Quantifiers would be able to specify default values for arrays/functions, but I don't think Yices support for quantifiers would handle the presence of the other theories (etc, for 0, 1, etc), and there must be some simpler way of handling default values for functions that doesn't get into quantifiers, perhaps along the lines above.
Do you happen to have pointers? @BrunoDutertre, any thoughts?
[...]

PR #633, opened 2026-05-27
smt2 array model display format

disteph:
## Summary

Fixes the model display part of #390. Symbols declared via the SMT-LIB `Array` sort are now
rendered in `(get-model)` as 0-arity array constants in SMT-LIB form,
e.g.

```smt2
(define-fun a () (Array Int Int)
  (store (store ((as const (Array Int Int)) 4) 2 5) 1 3))
```

instead of the previous, non-SMT-LIB-conformant function form

```smt2
(define-fun a ((x!0 Int)) Int (ite (= x!0 1) 3 (ite (= x!0 2) 5 4)))
```

which has both the wrong arity and the wrong sort relative to the
user's declaration. This is required for the SMT-COMP 2026 Model
Validation Track.

## Root cause

Once interned, Yices' type system represents both

```smt2
(declare-fun a () (Array K V))
(declare-fun f (K) V)
```

as an `UNINTERPRETED_TERM` of internal function type `(-> K V)`, so by
the time [print_smt2_model](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4060:0-4099:1) runs the SMT-LIB sort information is gone.
The fix records the missing bit (*"this name was declared as an
`Array` constant"*) in the SMT2 frontend at parse time and consults it
in the model printer.

## Implementation

Scoped entirely to [/Users/sgl/git/yices/yices2/src/frontend/smt2/](cci:4://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/:0:0-0:0).

- **`smt2_globals_t.array_const_terms`** — new `int_hmap_t` (term id →
  1 marker). Populated in [smt2_declare_fun](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:7002:0-7038:1) whenever the declaration
  is 0-arity and the resulting internal Yices type is a unary function
  type. Under SMT-LIB syntax that combination can only come from the
  `Array` sort (directly, via `define-sort`, or nested), so the check
  is both necessary and sufficient. `declare-const` is automatically
  covered because the SMT2 parser reduces it to `SMT2_DECLARE_FUN`.

- **Lifecycle.** Init/delete in [init_smt2_globals](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4676:0-4762:1) /
  [delete_smt2_globals](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4753:0-4811:1). `(pop)` is hooked via [smt2_pop_term_names](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:370:0-403:1),
  which now does a guarded `int_hmap_find` → `int_hmap_erase` per
  popped term (most popped names are not arrays, so the lookup avoids
  the cost of a no-op erase and the requirement that the record
  exists). `(reset-assertions)` calls `int_hmap_reset` alongside
  `yices_reset_tables()` only when `!global_decls`, since
  `yices_reset_tables` recycles term ids and stale markers would
  otherwise alias new unrelated terms; with `--global-decls` the map
  is preserved, matching the rest of the reset semantics. `(reset)` is
  covered automatically: [smt2_reset_all](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:7206:0-7232:1) destroys and re-inits the
  globals.

- **Printer.** [smt2_pp_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:755:0-788:1) gains a `bool array_const` parameter.
  When set, the definition is emitted as `(define-fun name () τ
  <value>)` regardless of the value's internal representation, and
  dispatches into the existing [smt2_pp_object_in_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:626:0-638:1) →
  [smt2_pp_array](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:516:0-562:1) / [smt2_pp_array_update](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:564:0-612:1) path which already produces
  `store` / `(as const τ)` syntax. [print_smt2_model](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4060:0-4099:1) looks up the
  marker via `int_hmap_find` and threads the flag through.

## Scope

- **In scope:** the SMT-LIB-style model output (`clean_model_format
  == false`), which is the default for the `yices_smt2` CLI binary
  ([smt2_force_smt2_model_format()](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4884:0-4889:1) at startup) and the configuration
  used by SMT-COMP. Library users obtain it the same way.
- **Out of scope:** the Yices-native model format
  (`clean_model_format == true`, `smt2_pp_full_model`) emits
  `(function ...)` / `(default ...)` syntax that is not valid SMT-LIB
  regardless of arrays vs functions. Fixing it requires rewriting that
  printer on top of [smt2_pp_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:755:0-788:1); tracked for a follow-up.
- **Out of scope:** accepting `(as const (Array K V))` as **input**.
  Issue #390 also asks for that direction; it requires changes to the
  parser and type system that are orthogonal to model output. This PR
  delivers the output direction; input support remains a separate
  feature.

## Manual verification

Confirmed correct rendering on:

- `(declare-fun a () (Array Int Int))`
- `(declare-const a (Array Int Int))`
- `(declare-fun a () (Array Int (Array Int Int)))` (nested)
- `(declare-fun a () (Array (_ BitVec 4) (_ BitVec 4)))`
- `(declare-fun a () (Array Int Real))` (Int → Real coercion in
  `(as const ...)`)
- (define-sort A () (Array Int Int)) + (declare-fun a () A)
- (declare-fun f (Int) (Array Int Int)) (UF with array range —
  array-of-array store/const at each leaf)
- Mixed array constant and plain UF in the same problem (UF must keep
  its ((x!0 K)) V (ite ...) form)
- (push 1) / array decl / (pop 1) / re-declare same name as a
  function (marker must not leak across pop)
- (reset-assertions) / re-declare same name as Int (marker must
  not survive term-id recycling)

## Regression tests

Six new gold-checked tests under [/Users/sgl/git/yices/yices2/tests/regress/](cci:4://file:///Users/sgl/git/yices/yices2/tests/regress/:0:0-0:0), matching the
existing issNNN_* convention:

- [iss390_array_basic.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/iss390_array_basic.smt2:0:0-0:0)
- [iss390_array_const.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/iss390_array_const.smt2:0:0-0:0)
- [iss390_array_mixed.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/iss390_array_mixed.smt2:0:0-0:0)
- [iss390_array_define_sort.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/iss390_array_define_sort.smt2:0:0-0:0)
- [iss390_array_pushpop.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/iss390_array_pushpop.smt2:0:0-0:0) (--incremental)
- [iss390_array_reset.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/iss390_array_reset.smt2:0:0-0:0) (--incremental)

All six pass under [tests/regress/check.sh](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/check.sh:0:0-0:0).

The full [/Users/sgl/git/yices/yices2/tests/regress](cci:4://file:///Users/sgl/git/yices/yices2/tests/regress:0:0-0:0) suite shows **no new regressions**: the same
245 pre-existing failures are present on master and on this branch
(predominantly wd/, coverage/random/QF_AUFBV/, and mcsat/ cases
that are known-broken before this work).

## Files touched

- [src/frontend/smt2/smt2_commands.h](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.h:0:0-0:0)
- [src/frontend/smt2/smt2_commands.c](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:0:0-0:0)
- [src/frontend/smt2/smt2_printer.h](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.h:0:0-0:0)
- [src/frontend/smt2/smt2_printer.c](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:0:0-0:0)
- tests/regress/iss390_array_*.smt2{,.gold,.options} (new)

### DIFF
--- a/src/frontend/smt2/smt2_commands.c
+++ b/src/frontend/smt2/smt2_commands.c
@@ -371,9 +371,15 @@ static void smt2_push_name(smt2_name_stack_t *s, char *name) {
 /*
  * Remove names on top of the stack and remove them from the term_name table
  * - ptr = new top: names[0 ... ptr-1] are kept
+ *
+ * Also drop any array-constant markers in __smt2_globals.array_const_terms
+ * for the popped term ids. Most popped names are not arrays, so we look up
+ * the marker first and only erase if present.
  */
 static void smt2_pop_term_names(smt2_name_stack_t *s, uint32_t ptr) {
   char *name;
+  term_t t;
+  int_hmap_pair_t *p;
   uint32_t n;
 
   n = s->top;
@@ -381,7 +387,13 @@ static void smt2_pop_term_names(smt2_name_stack_t *s, uint32_t ptr) {
     n --;
     name = s->names[n];
 
-    assert(yices_get_term_by_name(name) != NULL_TERM);
+    t = yices_get_term_by_name(name);
+    assert(t != NULL_TERM);
+    p = int_hmap_find(&__smt2_globals.array_const_terms, t);
+    if (p != NULL) {
+      int_hmap_erase(&__smt2_globals.array_const_terms, p);
+    }
+
     yices_remove_term_name(name);
     assert(yices_get_term_by_name(name) == NULL_TERM);
 
@@ -4078,7 +4090,9 @@ static void print_smt2_model(smt2_pp_t *printer, smt2_model_t *sm) {
      */
     if (good_object(vtbl, v)) {
       tau = term_type(terms, t);
-      smt2_pp_def(printer, vtbl, sm->names.data[i], tau, v);
+      bool array_const =
+        int_hmap_find(&__smt2_globals.array_const_terms, t) != NULL;
+      smt2_pp_def(printer, vtbl, sm->names.data[i], tau, v, array_const);
     }
   }
   pp_close_block(&printer->pp, true);
@@ -4721,6 +4735,7 @@ static void init_smt2_globals(smt2_globals_t *g) {
   init_named_term_stack(&g->named_asserts);
 
   init_pvector(&g->model_term_names, 0);
+  init_int_hmap(&g->array_const_terms, 0);
 
   g->unsat_core = NULL;
   g->unsat_assumptions = NULL;
@@ -4789,6 +4804,7 @@ static void delete_smt2_globals(smt2_globals_t *g) {
   delete_named_term_stack(&g->named_asserts);
 
   delete_string_vector(&g->model_term_names);
+  delete_int_hmap(&g->array_const_terms);
 
   if (g->unsat_core != NULL) {
     free_assumptions(g->unsat_core);
@@ -7020,6 +7036,20 @@ void smt2_declare_fun(const char *name, uint32_t n, type_t *tau) {
     save_term_name(&__smt2_globals, name);
     save_name_for_model(&__smt2_globals, name);
 
+    /*
+     * Record array-constant declarations so that the model printer can
+     * emit them in SMT-LIB store/const form rather than as a function
+     * with parameters. A 0-arity declaration whose internal Yices type
+     * is a unary function type can only have come from (Array K V) in
+     * the SMT-LIB grammar (directly, via define-sort, or nested).
+     */
+    if (n == 0
+        && is_function_type(__yices_globals.types, sigma)
+        && function_type_arity(__yices_globals.types, sigma) == 1) {
+      int_hmap_pair_t *p = int_hmap_get(&__smt2_globals.array_const_terms, t);
+      p->val = 1;
+    }
+
     report_success();
   }
 }
@@ -7188,10 +7218,14 @@ void smt2_reset_assertions(void) {
       ivector_reset(&g->val_vector);
 
       /*
-       * Reset the internal name tables, unless global_decls is set
+       * Reset the internal name tables, unless global_decls is set.
+       * yices_reset_tables() recycles term ids, so we must also drop
+       * stale array-constant markers; otherwise they could alias
+       * freshly minted unrelated terms after the reset.
        */
       if (!g->global_decls) {
         yices_reset_tables();
+        int_hmap_reset(&g->array_const_terms);
       }
 
       // build a fresh empty context
--- a/src/frontend/smt2/smt2_commands.h
+++ b/src/frontend/smt2/smt2_commands.h
@@ -47,6 +47,7 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#include "utils/int_hash_map.h"
 #include "utils/int_vectors.h"
 #include "utils/ptr_vectors.h"
 #include "utils/string_hash_map.h"
@@ -429,6 +430,15 @@ typedef struct smt2_globals_s {
   // This is used if clean_model_format is false to keep track of all
   // terms whose value we may need to print.
   pvector_t model_term_names;
+
+  // Map term_id -> 1 for term ids that were declared as SMT-LIB array
+  // constants (i.e., via (declare-fun s () (Array K V)) or
+  // (declare-const s (Array K V))). We record this here because once
+  // the term/type are interned, Yices' type system no longer
+  // distinguishes Array K V from a unary function. Used by
+  // print_smt2_model to emit (define-fun s () (Array K V) (store ...))
+  // instead of the function form (define-fun s ((x!0 K)) V (ite ...)).
+  int_hmap_t array_const_terms;
 
   // data structures for unsat cores/unsat assumptions
   // allocated on demand
--- a/src/frontend/smt2/smt2_printer.c
+++ b/src/frontend/smt2/smt2_printer.c
@@ -756,14 +756,25 @@ static void smt2_pp_update_definition(smt2_pp_t *printer, value_table_t *table, 
 /*
  * Print a definition in the SMT2 style.
  * - this prints (define-fun name () tau value)
+ *
+ * If array_const is true, the value is rendered as a 0-arity constant
+ * using store / (as const tau) regardless of the value's internal
+ * representation as a function or update. This matches the SMT-LIB
+ * model format for symbols declared via the Array sort.
  */
-void smt2_pp_def(smt2_pp_t *printer, value_table_t *table, const char *name, type_t tau, value_t c) {
+void smt2_pp_def(smt2_pp_t *printer, value_table_t *table, const char *name,
+                 type_t tau, value_t c, bool array_const) {
   type_table_t *types;
 
   types = table->type_table;
   pp_open_block(&printer->pp, PP_OPEN_SMT2_DEF);
   smt2_pp_symbol(printer, name);
-  if (object_is_function(table, c)) {
+  if (array_const) {
+    // (define-fun name () (Array K V) <store/const value>)
+    pp_string(&printer->pp, "()");
+    smt2_pp_type(printer, types, tau);
+    smt2_pp_object_in_def(printer, table, tau, c);
+  } else if (object_is_function(table, c)) {
     smt2_pp_function_params(printer, types, tau);
     smt2_pp_function_definition(printer, table, tau, c);
   } else if (object_is_update(table, c)) {
--- a/src/frontend/smt2/smt2_printer.h
+++ b/src/frontend/smt2/smt2_printer.h
@@ -125,9 +125,16 @@ extern void smt2_pp_queued_functions(smt2_pp_t *printer, value_table_t *table, b
  * Print a definition in the SMT2 style:
  *
  *   (define-fun name () tau value)
+ *
+ * If array_const is true, the definition is rendered as a 0-arity
+ * constant of array type, with the value built from store/(as const ...)
+ * instead of as a function with parameters and an if-then-else cascade.
+ * This is required for symbols originally declared via the SMT-LIB
+ * Array sort, since the model output must match the declaration's
+ * arity and sort.
  */
 extern void smt2_pp_def(smt2_pp_t *printer, value_table_t *table, const char *name,
-			type_t tau, value_t c);
+			type_t tau, value_t c, bool array_const);
 
 
 /*
--- a/tests/regress/iss390_array_basic.smt2
+++ b/tests/regress/iss390_array_basic.smt2
@@ -1,0 +1,10 @@
+; Issue #390: SMT-LIB Array-declared constants must be printed in the
+; model as 0-arity (define-fun name () (Array K V) (store ... (as const ...) ...))
+; rather than as a function (define-fun name ((x!0 K)) V (ite ...)).
+(set-logic QF_AUFLIA)
+(declare-fun a () (Array Int Int))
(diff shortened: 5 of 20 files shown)


disteph:
Review provided by Codex/GPT5.5:

## Findings

No findings.

## Open Questions / Assumptions

I reviewed `fix/smt2-array-model-format` against `master` at merge base `2137d2aafd5b1a849cb4be044a417f6cfed64c96`. The reviewed branch head is `35cb7d9584dea41447ed72f67c4df340e04c7373`.

## Summary

The change records which SMT-LIB declarations were array constants so model printing can emit 0-arity Array definitions using `store` / `(as const ...)`, while preserving normal unary-function model output. The lifecycle handling covers initialization, deletion, `pop`, and `reset-assertions`, and the tests cover basic arrays, `declare-const`, `define-sort`, mixed array/function declarations, push/pop, reset, MCSAT array model output, and existing `get-value` behavior.

## Tests Run

- `make clean`
- `make -j8`
- `tests/regress/run_test.sh tests/regress/iss390_array_basic.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/iss390_array_const.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/iss390_array_define_sort.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/iss390_array_mixed.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/iss390_array_pushpop.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/iss390_array_reset.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/mcsat/arrays/issue613_model_print.smt2 build/arm-apple-darwin23.6.0-release/bin`
- `tests/regress/run_test.sh tests/regress/wd/example_mdl.smt2 build/arm-apple-darwin23.6.0-release/bin`

disteph:
## Review by Windsurf/Opus4.7

I checked out the branch, walked the diff, and re-ran the new and updated regress tests against `master` and against this branch.

## Verifying the gold updates in `35cb7d95`

The third commit updates two pre-existing `.gold` files:

- [/Users/sgl/git/yices/yices2/tests/regress/mcsat/arrays/issue613_model_print.smt2.gold](cci:4://file:///Users/sgl/git/yices/yices2/tests/regress/mcsat/arrays/issue613_model_print.smt2.gold:0:0-0:0) — declares `a () (Array Int (Array Int Int))`. The previous gold had the `(get-model)` line in function form (`((x!0 Int)) (Array Int Int) (ite ...)`) while the very next `(get-value (a))` line on the same model already printed the value in `store`/`(as const ...)` form. So the old gold had the model and `get-value` outputs in two different syntaxes for the same term — the gold update removes that inconsistency.
- [/Users/sgl/git/yices/yices2/tests/regress/wd/example_mdl.smt2.gold](cci:4://file:///Users/sgl/git/yices/yices2/tests/regress/wd/example_mdl.smt2.gold:0:0-0:0) — same story for `(Array Int Bool)`. New gold lines up with the `get-value` block right below.

So both updates are correct, and the fact that the inconsistency existed in two different `.gold` files in two different subdirectories is a useful corroboration that the bug was real and that the fix is the right one.

Minor: it's worth mentioning these two gold updates explicitly in the PR description (currently they only show up via the file list). A reviewer skimming the diff may otherwise wonder why `mcsat/` and `wd/` `.gold` files moved.

## Marking heuristic

The criterion in [smt2_declare_fun](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:6989:0-7025:1) is:

```c
if (n == 0
    && is_function_type(__yices_globals.types, sigma)
    && function_type_arity(__yices_globals.types, sigma) == 1) { ... }
```

This is sound under the SMT-LIB 2 grammar: the only way to introduce a 0-arity term of internal unary-function type from the SMT2 frontend is via `(declare-fun a () (Array K V))` (or `(declare-const a (Array K V))`, or a `define-sort` macro that expands to `Array`). There is no SMT-LIB syntax for a non-`Array` unary function constant. The inline comment already says as much; I'd just suggest tightening it to call out the *grammar* invariant explicitly, e.g. *"Under SMT-LIB 2, a 0-arity declaration of unary function type can only originate from the Array sort"*, so a future contributor adding, say, a new theory with a non-Array unary-function constant doesn't accidentally trip this branch without realizing.

## Lifecycle hooks

- `init`/`delete` in [init_smt2_globals](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4678:0-4764:1) / [delete_smt2_globals](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4753:0-4811:1) — fine.
- `(pop)` is hooked inside [smt2_pop_term_names](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:370:0-391:1). The guarded `int_hmap_find` → `int_hmap_erase` is correct.
  - One nit: [smt2_pop_term_names](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:370:0-391:1) previously had no dependency on `__smt2_globals`, while its peers ([smt2_pop_type_names](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:406:0-427:1), [smt2_pop_macro_names](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:430:0-456:1)) still don't. The new direct reference to `__smt2_globals` mildly breaks symmetry. Not a blocker — and the alternative (call back from [smt2_pop](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:6621:0-6694:1) after the loop) would require either a parallel pre-pop scan or a second pass, which is uglier. Worth a one-line comment in the helper noting that this is the only term-stack helper that touches global state and why.
- `(reset-assertions)`: `int_hmap_reset` is gated on `!g->global_decls`, which mirrors the gate on `yices_reset_tables()`. Correct: with `--global-decls`, term ids are not recycled and the map must be preserved; without, ids are recycled and the map must be cleared.
- `(reset)`: [smt2_reset_all](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:7207:0-7233:1) already destroys + re-inits all globals, so it's covered automatically. Verified.

I don't see any other paths that could recycle term ids out from under the map (no internal Yices term GC fires during a normal SMT2 session). Good.

## API change to [smt2_pp_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:755:0-788:1)

[smt2_pp_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:755:0-788:1) is declared `extern` in [smt2_printer.h](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.h:0:0-0:0), so adding a `bool array_const` parameter is technically an ABI/API break. I `grep`ed the tree and there is exactly one caller ([print_smt2_model](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:4060:0-4099:1) in [smt2_commands.c](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_commands.c:0:0-0:0)), so this is fine in-tree, but if any downstream tool links against `libyices.a` and uses [smt2_pp_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:755:0-788:1) directly they'll need to update. The header docstring already explains the new parameter; I'd add one line in [/Users/sgl/git/yices/yices2/etc/NOTICES](cci:4://file:///Users/sgl/git/yices/yices2/etc/NOTICES:0:0-0:0) (or wherever ABI changes are tracked, if anywhere) just to be safe.

## Behavior on the `clean_model_format == true` path

The PR description correctly scopes itself to `clean_model_format == false`. To confirm: I traced `smt2_pp_full_model` ([smt2_model_printer.c](cci:7://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_model_printer.c:0:0-0:0)) and it still emits the Yices-native `(function ...)` / `(default ...)` syntax regardless of arrays. So the marker is silently ignored on that path, by design. As long as that's clearly stated in the changelog/release notes (and ideally in the man page or [/Users/sgl/git/yices/yices2/doc/manual/](cci:4://file:///Users/sgl/git/yices/yices2/doc/manual/:0:0-0:0)), this is fine. The follow-up to fix that path on top of [smt2_pp_def](cci:1://file:///Users/sgl/git/yices/yices2/src/frontend/smt2/smt2_printer.c:755:0-788:1) is the right plan.

## Tests

The six new `iss390_array_*` tests cover the documented matrix well. One small gap I'd suggest adding only if it's cheap: a `--global-decls` variant of `iss390_array_reset` that confirms the map is **preserved** across `(reset-assertions)` (i.e., the existing array marker still produces array-form output post-reset). It's the symmetric case to the existing test and protects the conditional gate on `!g->global_decls`.

Optionally, a Real-array test (`(Array Int Real)` with an integer default) would lock in the int→real coercion in the `(as const ...)` constant — already verified manually in the PR description, but not covered by a `.gold` test.

## Approval direction

LGTM modulo the small documentation/comment suggestions above. The implementation is minimal, the lifecycle reasoning is sound, the test coverage matches the manual matrix, and the two pre-existing `.gold` updates are independent corroborating evidence that the fix is correct.

ahmed-irfan (approved):

PR #634, opened 2026-05-29
mcsat: fall back to model cache when target cache value is infeasible

ahmed-irfan:
When deciding a value, the feasibility-aware plugins (na, ff, bv, uf) used a single cached hint (target cache preferred, else model cache). If that value was infeasible they went straight to a fresh pick, discarding the model cache's saved phase.

Try both cached tiers in priority order instead: tier 1 (target cache), then tier 2 (model cache), each checked against the variable's feasible set, before falling back to a fresh pick. The new behavior only fires when the target value is present but infeasible and the model value is present, feasible, and different - every other path is unchanged.

Add trail_get_cached_candidates() to own the priority ordering, NULL-stripping, and dedup (skip tier 2 when identical to tier 1) in one place; each plugin's decide path is now a plain loop over the returned candidates with only its own feasibility predicate. bool_plugin keeps the original combined getter (it has no feasible set).

Also fix a latent copy-paste assert in mcsat_value_eq (VALUE_LIBPOLY vs VALUE_RATIONAL comparison asserted v1->type instead of v2->type).

Full regression suite: 1726/1726 pass.

### DIFF
--- a/src/mcsat/bv/bv_feasible_set_db.c
+++ b/src/mcsat/bv/bv_feasible_set_db.c
@@ -222,16 +222,18 @@ const mcsat_value_t* bv_feasible_set_db_pick_value(bv_feasible_set_db_t* db, var
   term_t x_term = variable_db_get_term(db->ctx->var_db, x);
   uint32_t x_bitsize = term_bitsize(db->ctx->terms, x_term);
 
-  // Check the cached value from the
+  // Check the cached values: the trail returns the cached hints in priority order;
+  // return the first that is in the feasible set.
   const mcsat_trail_t* trail = db->ctx->trail;
-  if (trail_has_cached_value(trail, x)) {
-    const mcsat_value_t* cached_value = trail_get_cached_value(trail, x);
+  const mcsat_value_t* candidates[2];
+  uint32_t num_candidates = trail_get_cached_candidates(trail, x, candidates);
+  for (uint32_t i = 0; i < num_candidates; ++i) {
+    const mcsat_value_t* cached_value = candidates[i];
     if (x_feasible_full) {
       return cached_value;
-    } else {
-      ok = bv_bdd_manager_is_model(db->bddm, x_term, x_feasible_bdd, &cached_value->bv_value);
-      if (ok) { return cached_value; }
     }
+    ok = bv_bdd_manager_is_model(db->bddm, x_term, x_feasible_bdd, &cached_value->bv_value);
+    if (ok) { return cached_value; }
   }
 
   // Initialize the value we're using
--- a/src/mcsat/ff/ff_plugin.c
+++ b/src/mcsat/ff/ff_plugin.c
@@ -784,12 +784,15 @@ void ff_plugin_decide(plugin_t* plugin, variable_t x, trail_token_t* decide_toke
   const mcsat_value_t *x_new = NULL;
   mcsat_value_t x_new_local;
 
-  // see if there is a fitting cached value
-  if (trail_has_cached_value(ff->ctx->trail, x)) {
-    x_new = trail_get_cached_value(ff->ctx->trail, x);
-    assert(x_new->type == VALUE_LIBPOLY);
-    assert(x_new->lp_value.type == LP_VALUE_INTEGER);
-    if (feasible == NULL || lp_feasibility_set_int_contains(feasible, &x_new->lp_value.value.z)) {
+  // see if there is a fitting cached value: the trail returns the cached hints in
+  // priority order; check each against the feasible set and take the first fit.
+  const mcsat_value_t* x_candidates[2];
+  uint32_t x_num_candidates = trail_get_cached_candidates(ff->ctx->trail, x, x_candidates);
+  for (uint32_t i = 0; i < x_num_candidates && !using_cached; ++i) {
+    assert(x_candidates[i]->type == VALUE_LIBPOLY);
+    assert(x_candidates[i]->lp_value.type == LP_VALUE_INTEGER);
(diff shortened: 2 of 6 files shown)


disteph (approved):
# Code Review, with support from Windsurf/Opus4.8: PR #634 — mcsat: fall back to model cache when target cache value is infeasible

## Summary

The PR generalizes value-hint reuse in the feasibility-aware MCSAT plugins. Previously each plugin used a *single* cached hint ([trail_get_cached_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:212:0-222:1), which prefers the target cache and otherwise returns the model cache); if that one hint was infeasible the plugin discarded both and did a fresh pick. The new `trail_get_cached_candidates()` returns up to two hints in priority order (tier‑1 target cache, then tier‑2 model cache, deduped), and each plugin loops over them, applying its own feasibility predicate, before falling back to a fresh pick.

Overall this is a **clean, well-scoped change**. The new helper centralizes the priority/NULL-stripping/dedup logic, the per-plugin diffs are mechanical, and the behavior change is genuinely additive (only fires when tier‑1 is present-but-infeasible and tier‑2 is present, feasible, and distinct). I'd approve after a couple of clarifications below.

## Correctness

**Gating equivalence is preserved.** `trail_get_cached_candidates` returns `0` exactly when both caches are `VALUE_NONE`, which is the same condition as the old [trail_has_cached_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:176:0-183:1) guard. So plugins that dropped the explicit [trail_has_cached_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:176:0-183:1) call behave identically on the empty path.

**`x_feasible_full` path in bv is unchanged.** In `bv_feasible_set_db_pick_value`, when `x_feasible_full` the loop returns `candidates[0]` on the first iteration, which is tier‑1 == old [trail_get_cached_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:212:0-222:1). Same result.

**The [mcsat_value_eq](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/value.c:248:0-297:1) fix is a real latent bug, good catch.** In the `VALUE_LIBPOLY` vs non-LIBPOLY branch, `v1->type` is necessarily `VALUE_LIBPOLY`, so the old `assert(v1->type == VALUE_RATIONAL)` would always fire in debug builds if that branch were ever reached — the correct invariant is on `v2`. Worth noting in the commit that this implies the mixed LIBPOLY/RATIONAL path was effectively never exercised under assertions before; the new dedup [mcsat_value_eq(tier2, out[0])](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/value.c:248:0-297:1) makes it more important that this is right (although tier1/tier2 for the same var should share a type).

**UF plugin restructuring is correct but relies on an implicit invariant.** The new code builds the `forbidden` vector only on the *first* `eq_graph_get_forbidden` call and passes `values = NULL` for subsequent candidates, testing each against the already-built list. This is valid because `eq_graph_get_forbidden`'s `values` output is independent of `v` — `v` only affects the `different` return. The added comment documents this, which is good, but it's a real coupling: if `eq_graph_get_forbidden` ever made the list depend on `v`, this loop would silently break. Consider a one-line assertion or a note at `eq_graph_get_forbidden`'s declaration to lock the invariant.

I traced the UF candidate counts:

- **0 candidates**: first call passes `NULL` → builds list, `cache_ok=false`; loop `i=1..0` doesn't run. Matches old fresh-pick path.
- **1 candidate**: tests `candidates[0]`; loop doesn't run. Matches old single-hint behavior.
- **2 candidates**: tests `candidates[0]`; if infeasible, tests `candidates[1]` against the built list. New behavior, correct.

## Minor points

**ff_plugin now asserts the type invariant on both tiers.** The asserts `type == VALUE_LIBPOLY` / `lp_value.type == LP_VALUE_INTEGER` now apply to the tier‑2 (model cache) value as well, not just the single old hint. This should hold (the model cache stores prior ff assignments), but it widens the assertion surface — worth a sanity check that the model cache can never hold a non-integer libpoly value for an ff variable (e.g. transiently during construction).

**The `[2]` array size is duplicated across four call sites.** If a third tier is ever added, every `candidates[2]` / `out[2]` declaration and the helper need updating in lockstep. A named constant (e.g. `MCSAT_MAX_CACHED_CANDIDATES`) used by both the helper signature and the call sites would make that safer. Optional.

**Helper doc comment is accurate and helpful.** The "each value is only a hint; the caller must still check feasibility" note in [trail.h](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:0:0-0:0) is exactly the right thing to emphasize, since correctness depends entirely on the per-plugin feasibility check, not the cache.

## Testing

The PR reports full regression pass (1726/1726). Because the new path only triggers when tier‑1 is infeasible *and* tier‑2 is feasible *and* distinct, it's plausible existing regressions rarely hit it. If easy, a targeted note (or a small instrumentation count during a regression run) showing the tier‑2 fallback actually fires on some benchmarks would strengthen confidence that the new branch is exercised — otherwise the change is effectively dormant.

## Verdict

**Approve with minor comments.** Sound, well-factored refactor; the only correctness-relevant subtlety (UF's reuse of the prebuilt forbidden list) is real but currently valid and documented. The [mcsat_value_eq](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/value.c:248:0-297:1) assert fix is an independent, correct improvement worth keeping.

PR #635, opened 2026-06-01
MCSAT: L2O

ahmed-irfan:
Work done by 
- @elipparini
-  @Ovascos


disteph:
# Regression analysis for the CI failure

By Windsurf/Opus4.8.
Root cause: it's an *interaction* between merged-in #634 and your branch's #635 recache change

## The merge pulled in exactly one master commit

`git log 89535223^1..89535223^2` shows the merge `89535223` introduced a single commit from master:

- `7206d476` -- **PR #634, "mcsat: fall back to model cache when target cache value is infeasible"** (the PR I reviewed earlier).

So the only master change that could have triggered this is #634.

## What #634 changed (the BV decision value)

In `bv_feasible_set_db_pick_value` ([/Users/sgl/git/yices/yices2/src/mcsat/bv/bv_feasible_set_db.c:222](cci:4://file:///Users/sgl/git/yices/yices2/src/mcsat/bv/bv_feasible_set_db.c:221:0-221:999)):

- **Before #634:** take a *single* cached hint ([trail_get_cached_value](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:212:0-222:1) = target cache, else model cache). If it's infeasible, fall straight through to a **fresh pick**.
- **After #634:** try **both** tiers -- tier-1 `target_cache`, then tier-2 `model` cache -- and decide the first feasible one before any fresh pick.

Net effect: in the case "tier-1 present but infeasible, tier-2 present and feasible," the **decided BV value changed** from a freshly computed value to the model-cache value. Different decision => different MCSAT search trajectory.

## Why it only breaks *after* the merge (the #635 side)

#634 passed on master, and #635 passed at `2d347fb6`, yet the combination fails. That's because #635 itself changed the **recache scheduling** -- and recaching is exactly what populates the `target_cache` that #634's new tier-1/tier-2 logic reads. From `git diff 7206d476 89535223 -- src/mcsat/solver.c`:

```c
// REMOVED (old cadence, fired in the restart else-if chain, any trail level):
} else if (conflicts > recache_limit) {
  ++recache_round; mcsat_request_recache(mcsat); ...
}

// ADDED by #635 (now a separate check, gated on base level):
if (trail_is_at_base_level(mcsat->trail) && conflicts > recache_limit) {
  ++recache_round; mcsat_request_recache(mcsat); ...
}
```

Plus the initial delay change at [/Users/sgl/git/yices/yices2/src/mcsat/solver.c:380-381](cci:4://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:379:0-380:999). These are **not** L2O-gated, so they affect `QF_BV` too. `mcsat_request_recache` -> `trail_recache(...)` refreshes `trail->target_cache`, i.e. tier-1 in #634's [trail_get_cached_candidates](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:224:0-245:1).

So #635 changed **when/whether the target cache is refreshed**, and #634 changed **how the (now differently-populated) target/model caches drive BV decisions**.

## The truth table

- **#635 alone** (`2d347fb6`): new base-level recache cadence, but *old* single-hint BV selection -> no tier-2 fallback -> **passes**.
- **#634 alone** (master): new tier-2 fallback, but *old* recache cadence -> caches as before -> **passes**.
- **#634 + #635** (`89535223`): new cadence feeds a different `target_cache` *into* the new tier-2 fallback -> an emergent decision sequence on this `simple_processors_002_002_0008` instance -> pathological search.

This is why it surfaces only in the slow **debug + thread-safety + gcc/Linux** config: the altered decision order produces a much longer search, and that config runs each step ~10-30x slower (debug + assertions) with the two `mcsat::bv::*::check` traces adding per-conflict verification -- so a trajectory that's tolerable elsewhere can't finish within any practical CPU limit. Locally (release-ish/non-TS) the same input still solves in 0.44s.

## Implications for the fix

A timeout bump can't reliably fix an interaction that changes the search trajectory into a near-unbounded one. The robust options are unchanged from before -- most surgical is to **exclude this test in the debug+TS+mcsat config**. But if you'd rather treat the interaction as a real heuristic regression, the better fix is on the **#635 side**: the recache cadence change (base-level gating / initial delay) is what destabilizes #634's cache-driven BV decisions, so reconsidering that change (or making #634's tier-2 fallback insensitive to recache timing) would address the cause rather than the symptom.

disteph (commented):
# Review by Windsurf/Opus4.8: PR #635 — MCSAT: L2O

## What it does

Adds a "Learning-to-Optimize" (L2O) component to MCSAT: a numerical **hill-climbing local search** over a soft-cost encoding of the assertions that produces **decision-value hints** seeded into the trail's model cache. New module `src/mcsat/l2o/` ([l2o.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:0:0-0:0), [evaluator.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/evaluator.c:0:0-0:0), [hill_climbing.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/hill_climbing.c:0:0-0:0), [l2o_cost_fx.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o_cost_fx.c:0:0-0:0)), a new `int_hash_mmap` multimap util, a `mcsat-l2o` option, term/term-manager helpers exported for reuse, and two NRA regressions. It is wired into the recache cycle (`mcsat_process_requests`) and auto-enabled for `QF_NIA`.

## Soundness — the key reassurance

The design is **sound by construction**: L2O only ever writes *advisory hints* via [hint_value_to_trail](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:350:0-364:1), and only for currently-unassigned variables:

```c
// src/mcsat/l2o/l2o.c:362-364
  if(! trail_has_value(trail, var) ){
    mcsat_model_set_value(&trail->model, var, val);
  }
```

The decision plugins still validate every value against the feasible set before committing. So a bad/garbage L2O result can only cost time, never correctness. Termination is bounded (`MAX_ITER`/`MAX_CALLS` in [hill_climbing](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/hill_climbing.c:228:0-301:1), finite assertion/var sets). Good.

## High-impact concerns

### 1. The recache-cadence change is NOT L2O-gated and already caused a regression

In `mcsat_solve`, the recache trigger was rewritten (`src/mcsat/solver.c:3072`): it moved out of the restart `else-if` chain into a separate check **gated on [trail_is_at_base_level](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/trail.h:193:0-197:1)**, and the seed changed from `recache_interval` to `recache_initial_delay`. These apply to **every MCSAT solve, including QF_BV/QF_NRA with `l2o=false`** — they are not behind the `l2o` flag. This is exactly what destabilized [simple_processors_002_002_0008.smt2](cci:7://file:///Users/sgl/git/yices/yices2/tests/regress/mcsat/bv/simple_processors_002_002_0008.smt2:0:0-0:0) (QF_BV) once it combined with #634's cache-driven decision change. **Recommendation:** either gate the cadence change behind `mcsat_options.l2o`, or treat it as an independent heuristic change with its own benchmarking and PR, rather than shipping it inside an "L2O" feature where it silently affects unrelated logics.

### 2. Auto-enabling L2O for all `QF_NIA` by default

`smt2_set_logic` flips the feature on for `QF_NIA` unconditionally. That turns an experimental search on by default for a whole logic class. Confirm this is intended; if it's meant for evaluation, consider keeping it opt-in (or clearly documenting the default flip), and note it only catches `QF_NIA` (not `NIA`/`QF_NRA`).

### 3. [get_composite](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/terms.c:3533:0-3626:1) promoted to a public API while keeping a shared `static` scratch buffer

The relocation of [get_composite](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/terms.c:3533:0-3626:1)/`mk_composite` from [preprocessor.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:0:0-0:0) into [terms.c](cci:7://file:///Users/sgl/git/yices/yices2/src/terms/terms.c:0:0-0:0)/`term_manager.c` is a reasonable refactor, but [get_composite](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/terms.c:3533:0-3626:1) returns a pointer into a **single file-scope mutable buffer** for non-composite atoms:

```c
// src/terms/terms.c:3531-3532
static
composite_term1_t composite_for_noncomposite;
```

This was tolerable as a [preprocessor.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/preprocessor.c:0:0-0:0)-local helper, but it's now exported via `terms.h` and used more widely (including [l2o_calculate](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:199:0-334:1)). Two hazards: (a) the returned pointer is only valid **until the next [get_composite](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/terms.c:3533:0-3626:1) call** on a non-composite atom — easy to misuse when holding two descriptors; (b) it is **not thread-safe**, which matters given the `--enable-thread-safety` build. At minimum document the lifetime contract; better, have the caller pass storage (or use a thread-local).

## Medium

- **`get_fs_by_term` coupling.** It's a non-static global in `na_plugin.c` declared via a bare `extern` inside [l2o.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:0:0-0:0)/[hill_climbing.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/hill_climbing.c:0:0-0:0) rather than a header (`src/mcsat/l2o/l2o.c:553`). Put the prototype in an `na_plugin` header and give it an `na_`/`mcsat_` prefix. This is the concrete form of the author's own acknowledged "ugly hack" of stashing `na_plugin`/`bool_plugin` pointers in `l2o_t` (`src/mcsat/l2o/l2o.h:36-38`).

- **Dead declaration.** `trail_model_cache_clear` is declared in `src/mcsat/trail.h:329` but has no definition anywhere. Remove it (or implement if intended).

- **Recursive [collect_free_vars](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:80:0-187:1).** `src/mcsat/l2o/l2o.c:82` recurses on term structure, while [l2o_term_is_supported](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:790:0-875:1) deliberately uses an explicit stack. Deeply nested terms could overflow the stack; consider making [collect_free_vars](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:80:0-187:1) iterative too for consistency/robustness.

- **Duplicate include.** `#include <math.h>` appears twice in [solver.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/solver.c:0:0-0:0) (`src/mcsat/solver.c:66`).

## Minor / style

- Several `for (int i = 0; i < n; ...)` loops compare signed `int` against `uint32_t`/`fs->size` (e.g. [l2o_calculate](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:199:0-334:1) OR / [optimize_fs](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/hill_climbing.c:178:0-222:1)) — signed/unsigned comparison; use `uint32_t`.
- Commented-out scaffolding left in `mcsat_solve` (the `run_l2o`/[l2o_run](cci:1://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o.c:901:0-910:1) TODO blocks at `src/mcsat/solver.c:3117`) and a checked-in `src/mcsat/l2o/TODOs.txt` — consider trimming before merge.
- [l2o_cost_fx.c](cci:7://file:///Users/sgl/git/yices/yices2/src/mcsat/l2o/l2o_cost_fx.c:0:0-0:0) is missing the GPL license header that every other file carries.

## Tests

Two NRA regressions are added, which is good. But given the headline integration is `QF_NIA` auto-enable and the cost encoding handles bool/int/real, I'd want at least one `QF_NIA` regression that actually exercises the L2O path, plus a note that the suite was run with `--enable-thread-safety` (the config where the cache interaction bit us). The cost-function semantics (the `L2O_TRUE=0.0`/`L2O_FALSE=1.0` soft encoding, product-for-OR / sum-for-AND) are heuristic and only affect performance, so they don't need golden tests, but a targeted check that hints actually get consumed would raise confidence the feature isn't dormant.

## Verdict

**Request changes — but mostly for *scoping*, not soundness.** The L2O machinery itself is cleanly isolated and can't affect correctness. The blocking items are the two changes that leak outside the feature flag: the **non-gated recache-cadence change** (which already broke an unrelated QF_BV test) and the **default `QF_NIA` enable**. Gate or separately justify those, fix the [get_composite](cci:1://file:///Users/sgl/git/yices/yices2/src/terms/terms.c:3533:0-3626:1) lifetime/thread-safety exposure, and clean up the dead `trail_model_cache_clear` decl + stray includes/TODOs. After that it's a reasonable experimental addition.

disteph:
I'd suggest to address:
- get_composite promoted to a public API while keeping a shared static scratch buffer
- Dead declaration. trail_model_cache_clear
- Duplicate include. #include <math.h>
- Several for (int i = 0; i < n; ...) loops compare signed int against uint32_t/fs->size
- l2o_cost_fx.c is missing the GPL license header
- A couple of QF_NIA regressions that actually exercises the L2O path

disteph:
Looks good to me now!

PR #643, opened 2026-06-15
Wide model projection with implicant cube enumeration and arithmetic preprocessing

disteph:
This PR refines the existing SAT-guided `YICES_GEN_BY_PROJ_WIDE` model-generalization path, exposes implicant cube enumeration through the C API, and adds arithmetic preprocessing before projection.

The main goal is to make projection more robust on formulas containing arithmetic constructs that are not directly supported by the projection procedures, while preserving as much Presburger structure as possible.

## Projection Pipeline

For projection-based model generalization, the pipeline is now:

1. Rewrite `abs(t)` as the equivalent ITE term `ite(t >= 0, t, -t)`.

2. Build projection cubes:
   - `YICES_GEN_BY_PROJ` calls `get_implicant` once and uses the resulting normalized literal cube.
   - `YICES_GEN_BY_PROJ_WIDE` enumerates Boolean implicant cubes using the existing SAT-guided WIDE machinery.

3. For WIDE only, normalize each enumerated cube through `get_implicant` before projection.
   This consumes lingering model-irrelevant ITE branches, including ITEs introduced by the ABS rewrite when they survive cube enumeration.

4. Preprocess each normalized cube before projection:
   - eliminate `floor`, `ceil`, `idiv`, and `mod` where possible,
   - eliminate `rdiv` where possible.

5. Project each preprocessed cube.

6. For WIDE projection, union successfully projected cubes and use local fallback behavior when the cube budget is exhausted.

## Public implicant cube enumeration API

Adds:

- `yices_implicant_cubes_for_formula(...)`
- `yices_implicant_cubes_for_formulas(...)`

These return model-true implicant cubes for one formula or for an array of formulas.

The implementation reuses the existing SAT-guided Boolean abstraction machinery used by WIDE projection.

## WIDE projection refinements

Refines the existing `YICES_GEN_BY_PROJ_WIDE` path:

- `cube_budget` now caps the number of distinct normalized cubes attempted for projection.
- `cube_budget == 0` means unbounded.
- normalized cubes are deduplicated before projection attempts,
- per-cube projection failures skip only that cube,
- budget exhaustion can return:
  - the collected projected cubes,
  - a local projection fallback,
  - or `OR(collected, local)`.

This avoids spending budget repeatedly on Boolean models that normalize to the same arithmetic cube.

## ABS and ITE Handling

The existing WIDE pipeline still relies on Boolean abstraction and implicant extraction to eliminate model-irrelevant ITE branches before projection.

This PR preserves that structure:

- ITEs are not eliminated by the arithmetic preprocessing pass itself.
- `abs(t)` is rewritten before cube enumeration as the equivalent term `ite(t >= 0, t, -t)`.
- When the ITE expansion/cube enumeration budget is not exhausted, these ABS-introduced ITEs are consumed by cube enumeration.
- If budget is exhausted and a cube still contains an ITE, the existing per-cube `get_implicant` normalization handles lingering model-relevant ITEs before projection.

The ITE expansion budget is tied to `cube_budget`; `cube_budget == 0` keeps both unbounded.

## Floor, Ceil, IDIV, and MOD Preprocessing

Adds cached per-cube preprocessing for:

- `floor`
- `ceil`
- `idiv`
- `mod`

The rewrite introduces fresh integer auxiliary variables whose model values match the construct being replaced. These auxiliary variables are passed as additional eliminands to projection.

This is intentionally per-cube rather than global:

- it avoids polarity/equivalence complications in arbitrary formula positions,
- it benefits from atom-level caching across cubes,
- it lets unsupported constructs in unenumerated cubes remain harmless.

Unsafe cases do not abort preprocessing. For example, if the divisor of `idiv` or `mod` evaluates to zero, the original expression is left unchanged. If that construct later reaches projection and is unsupported, only the affected cube is skipped.

## RDIV Preprocessing

Adds `src/model/projection_preprocess.{c,h}`.

After floor/ceil/idiv/mod preprocessing and before projecting a cube, RDIV-containing arithmetic literals are rewritten by:

- collecting each side of the atom into a single fraction,
- multiplying both sides by a common denominator,
- adding model-sign guards for denominators,
- switching inequality direction when the model sign requires it.

RDIV preprocessing is cached across cubes, so repeated atoms in different cubes reuse the same rewrite.

## Arithmetic Projection Routing and Salvage

Improves projection routing for arithmetic cubes that are not immediately accepted by a projection procedure.

The projection path now distinguishes more carefully between:

- real eliminands,
- integer eliminands,
- Presburger-compatible literals,
- non-Presburger arithmetic structure,
- divisibility atoms.

For mixed or initially unsupported arithmetic cubes:

- real eliminands may be substituted by their model values when needed to remove non-Presburger real structure,
- integer eliminands are substituted incrementally, one at a time, only as needed,
- after each substitution step, the cube is checked again for Presburger eligibility,
- divisibility atoms are allowed to survive these checks when they are Presburger-compatible,
- once the remaining cube is Presburger, the Presburger projection procedure is used.

This fixes the previous behavior where integer variables in NIA/NIRA cubes could be substituted too eagerly, losing useful Presburger constraints before the Presburger projection path had a chance to handle them.

## Affected Modes

The changes affect projection-based model generalization:

- `YICES_GEN_BY_PROJ`
- `YICES_GEN_BY_PROJ_WIDE`

`YICES_GEN_BY_PROJ_WIDE` additionally benefits from the cube enumeration refinements.

The non-wide projection path also benefits from RDIV preprocessing, arithmetic construct preprocessing, and the improved arithmetic projection routing.

`YICES_GEN_BY_SUBST` is not intended to change.

## Tests

Expanded API tests cover:

- public implicant cube enumeration,
- subset-minimal cube behavior,
- arithmetic ITE cube extraction,
- WIDE projection behavior,
- WIDE cube budgets,
- shared eliminands,
- polarity-sensitive projection cases,
- LIA Presburger divisibility,
- divisibility atoms reaching Presburger projection,
- NIA integer projection salvage,
- mixed integer/real arithmetic projection,
- ABS preprocessing through ITE introduction,
- floor/ceil preprocessing,
- idiv preprocessing,
- imod preprocessing,
- RDIV preprocessing with positive and negative denominators,
- RDIV hidden under arithmetic structure,
- RDIV in dead ITE branches.

## Verification

Run on the branch:

- `make -j4`
- `build/arm-apple-darwin23.6.0-release/bin/test_generalize_model`
- `build/arm-apple-darwin23.6.0-release/bin/issue_414_mcsat`

Debug rebuild from clean binaries:

- `make MODE=debug clean`
- `make MODE=debug -j4`
- `build/arm-apple-darwin23.6.0-debug/bin/test_generalize_model`
- `build/arm-apple-darwin23.6.0-debug/bin/issue_414_mcsat`

Also checked the standalone tiny QF_NIA MCSAT formula:

    (set-logic QF_NIA)
    (declare-fun x () Int)
    (declare-fun y () Int)
    (assert (= x (* y y)))
    (assert (= y 2))
    (check-sat)

with:

- `build/arm-apple-darwin23.6.0-debug/bin/yices_smt2 --mcsat --timeout=5 /tmp/yices_nia_probe.smt2`

Result:

    sat


disteph:
# Code Review by Windsurf/Opus4.8: PR #643 — Wide model projection with implicant cube enumeration and arithmetic preprocessing

## Summary

Large, well-structured PR (~4.7k insertions / ~317 deletions across 11 files). Three logically separable pieces:

1. **Cube enumeration rework** (`generalization.c`): replaces the `new_sat_solver`-based WIDE enumeration with an `smt_core` bool-only solver driven with strict false-first decisions + superset blockers to produce *subset-minimal* implicant cubes. Exposed via new public API `yices_implicant_cubes_for_formula[s]`.
2. **Arithmetic preprocessing** (new `projection_preprocess.{c,h}`): ABS→ITE rewrite, per-cube floor/ceil/idiv/mod elimination with fresh integer auxiliaries, and RDIV elimination via fraction cross-multiplication with model-sign guards.
3. **Projection routing/salvage** (`projection.c`): type-directed splitting of integer vs. real eliminands, incremental integer substitution to preserve Presburger structure, and an OR-of-routes fallback for mixed cubes.

Overall the design is sound and the "non-failing, skip-the-cube" philosophy is applied consistently, which keeps the soundness burden low. Comments below are mostly correctness-adjacent concerns, a few API/maintainability points, and suggested tests.

## Correctness / Soundness

### 1. ITE-expansion budget is silently coupled to `max_cubes` (public API smell)
In `enumerate_implicant_cubes_with_status`:
```c
init_abs_builder(&builder, mdl, mngr, &eval, max_cubes); // max_cubes used as max_ite_expansions
```
For the public `yices_implicant_cubes_for_formula(mdl, t, max_cubes, v)`, the user's *cube cap* doubles as the *ITE-expansion cap*. Consequences:
- `max_cubes = 1` silently produces a **coarser abstraction** (≤1 ITE expansion), not just fewer cubes — surprising for a caller that just wants one cube.
- `max_cubes = 0` (unbounded) means **unbounded ITE expansion**, which can blow up on deeply nested ITE / ABS-introduced ITE terms.

This coupling is reasonable internally for WIDE (and is described in the PR text), but it is **not documented in `yices.h`** for the public entry points. Recommend either documenting it explicitly or decoupling the ITE budget from the public `max_cubes`.

### 2. RDIV / sign guards can be nonlinear (effectiveness, not unsoundness)
[fraction_of_rdiv](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1751:0-1797:1) / [rewrite_arith_atom](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1880:0-1922:1) build a single `num/den` fraction and guard `den`'s model sign via `mk_arith_term_gt0(den)` / `lt0`. When `den` is a product of variables, the guard literal is nonlinear. The rewrite itself is **sound** (sign fixed to the model sign), but in non-MCSAT builds the nonlinear guard will make projection return `PROJ_ERROR_NON_LINEAR`, silently skipping the cube. This is safe but may quietly reduce WIDE coverage. Worth a comment near [model_sign_and_guard_for](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1562:0-1604:1) so future readers don't mistake skipped cubes for a bug.

### 3. idiv/mod and floor/ceil cell guards — checked, look sound
- [append_division_cell_guards](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1108:0-1139:1): for `sign POS`, `aux*rhs <= lhs < (aux+1)*rhs`; for `sign NEG`, `aux*rhs <= lhs < (aux-1)*rhs`. These match SMT-LIB Euclidean `div` for positive and negative divisors respectively, with the divisor sign pinned by `sign_guard`. Good.
- [append_floor_guards](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1046:0-1073:1): `aux <= arg < aux+1` (floor) / `arg <= aux < arg+1` form (ceil). Correct.
- `mod` reconstructed as `lhs - rhs*aux`. Correct given the quotient cell.

One thing to confirm: the `aux` model value comes from `mk_arith_idiv(lhs,rhs)` / `mk_arith_floor` evaluated in the model. Good that [quotient_model_value](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1141:0-1159:1) bails (returns false → literal left unchanged) when `sign == ZERO`, so div/mod-by-zero is left to projection to reject. Sound.

### 4. Permanent term-table growth from fresh auxiliaries
[fresh_int_aux](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:848:0-868:1) calls `new_uninterpreted_term(...)` for each floor/ceil/idiv/mod atom (cached per atom). On cache deletion only the **model mappings** are erased (`int_hmap_erase(&mdl->map, ...)`); the **terms themselves persist** in the global term table (append-only). For one generalization call this is bounded, but across many calls / large WIDE runs with many distinct atoms it is a slow monotonic growth of the term table. Likely acceptable given Yices never GCs terms, but worth a one-line note in [fresh_int_aux](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:848:0-868:1).

### 5. All-NULL theory SMT interface
```c
static th_smt_interface_t gen_bool_smt = { NULL, NULL, NULL, NULL, NULL };
```
This relies entirely on `smt_core_set_bool_only(core)` guaranteeing these hooks are never invoked. If that invariant ever changes, a NULL call crashes hard. Consider either pointing them at no-op stubs (as done for `gen_bool_ctrl`) or adding an assertion/comment documenting the bool-only precondition.

### 6. Mixed int/real route OR — the riskiest new soundness logic
`proj_process_arith_literals` runs **two full projector copies** (Presburger-salvage and real-arith), then `proj_commit_route_or` factors common conjuncts and ORs the residuals. Correctness depends on each route independently producing a valid generalization and on `proj_factor_common_route_terms` only factoring literals that are truly shared. This is intricate; the logic reads correctly to me, but it is the place most likely to harbor a subtle over-generalization. See test suggestions below.

### 7. `assert(code > 0)` in `gen_model_by_proj_sat_guided`
After enumeration, `assert(code > 0)` then `num_cubes = (uint32_t) code`. I traced the enumerator and `num_cubes >= 1` does hold on all non-negative-return paths (n==0, root true, local fallback, and ≥1 loop iteration since the budget check can't trip on the first iteration). So the assert is valid, and the NDEBUG path degrades gracefully (have_first stays false → local fallback). Fine, but since it's effectively an invariant on a separate function's return, a short comment justifying it would help.

## API / Documentation

- `yices.h` cube-stream contract (k cubes, k−1 `NULL_TERM` separators, no trailing separator, empty vector ⇒ one `true` cube) is clearly written and matches [test_implicant_cubes.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_implicant_cubes.c:0:0-0:0). 
- Please add the `max_cubes`↔ITE-expansion coupling note (point #1) to the public doc.
- The updated WIDE budget semantics in `generalization.h` / `yices.h` (distinct normalized cubes, dedup, union-only on budget exhaustion) are consistent across header, impl comments, and PR description. 

## Maintainability / Minor

- `replace_ite_subterm` + the parallel `find_ite_subterm_*` + `rewrite_abs_*` + `rewrite_construct_*` + `term_contains_*` families are each large per-term-kind switches that repeat the same `safe_malloc`/recurse/`safe_free` pattern. Heavy but readable. The repeated `composite_term_desc` fan-outs in [term_contains_arith_construct](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:897:0-978:1) and [term_contains_rdiv](cci:1://file:///Users/sgl/git/yices/yices2/src/model/projection_preprocess.c:1453:0-1536:1) are near-duplicates; a shared "any-child-matches" helper would cut ~80 lines and reduce drift risk.
- `tests/api/Makefile` uses `$(wildcard *.c)`, so [test_implicant_cubes.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_implicant_cubes.c:0:0-0:0) is picked up automatically — good, no Makefile edit needed.
- `gen_bool_backtrack`/`gen_bool_donothing` unused-parameter warnings: confirm `-Wunused-parameter` is not in the strict set, or add `(void)` casts (you already do this for `mdl` in `init_abs_builder`).
- `proj_init_route_copy` copies vectors but I'd double-check `delete_projector` on the route copies frees everything `init_projector` allocated (esp. `avars_to_keep` / `arith_proj` / `presburger`) to avoid a leak per mixed cube.

## Tests

Good breadth already (ABS→ITE, floor/ceil/idiv/imod, RDIV with ±denominators and hidden/dead-branch RDIV, NIA salvage, mixed int/real, budgets, subset-minimality). Suggested additions targeting the riskiest paths:

- **Route-OR soundness**: a mixed NIRA cube where Presburger-salvage and real-arith routes produce *different* residuals, asserting the resulting cell still implies the target (the [assert_cube_contract](cci:1://file:///Users/sgl/git/yices/yices2/tests/api/test_implicant_cubes.c:53:0-67:1) pattern from [test_implicant_cubes.c](cci:7://file:///Users/sgl/git/yices/yices2/tests/api/test_implicant_cubes.c:0:0-0:0) is ideal here).
- **idiv/mod by negative divisor** reaching projection (verify the `sign NEG` cell guards), plus **div/mod-by-zero in a dead/irrelevant literal** (should leave literal unchanged, not error).
- **Nonlinear RDIV sign guard** in a non-MCSAT build: confirm the cube is skipped (not unsound, not aborting) and WIDE still returns a valid result from other cubes.
- A **deeply nested ITE/ABS** input with small `max_cubes` to lock in the budget-coupling behavior (whatever you decide for #1).

## Verdict

Architecturally solid and the conservative skip-on-failure design keeps soundness risk contained. I'd consider it close to mergeable once:
- the public `max_cubes`/ITE-expansion coupling is documented or decoupled (#1),
- the route-OR path gets a dedicated implies-target test (#6),
- and the small comments (#2, #4, #5, #7) are added.

I reviewed the diff statically and did **not** rebuild; recommend a clean `make MODE=debug clean && make MODE=debug` plus the listed `test_generalize_model` / `test_implicant_cubes` runs (under `timeout`) before merge.

disteph:
1. We kept the ITE-expansion behavior internal and documented only the public effect: `max_cubes`/`cube_budget` control how much work is attempted, larger values make calls more expensive, and returned implicant literals never contain ITEs.

2. We added a comment near RDIV sign-guard construction explaining that nonlinear guards are expected and may cause non-MCSAT projection to skip a cube without compromising soundness.

3. We left the floor/ceil/idiv/mod cell-guard logic unchanged, since the review confirmed it as sound and the div/mod-by-zero path already leaves literals unchanged instead of failing preprocessing.

4. We added a comment in `fresh_int_aux` clarifying that deleting the cache removes temporary model mappings, not the fresh terms, because Yices term tables are append-only.

5. We added a comment on the all-NULL SMT theory interface documenting that the solver is put in bool-only mode and these hooks must be unreachable.

6. We added a mixed integer/real route-OR regression test showing Presburger structure survives route combination and the resulting generalization admits even values while rejecting an odd one.

7. We added a comment before `assert(code > 0)` explaining the enumerator’s success invariant: a model-true formula has at least one encoded cube, possibly the empty cube.

PR #644, opened 2026-06-23
2.8.0 doc updates

disteph:
This is the branch for updating the doc to 2.8.0.


ahmed-irfan (commented):
Thanks for the doc update — the finite-field coverage and error-code additions look good. A few things to check before merging:
  
  Possibly missing function docs

  Four public API functions appear in yices.h but have no .. c:function:: entry in any RST file on master. Please confirm whether
  this PR documents them, or add entries if not:

  - yices_implicant_cubes_for_formula / yices_implicant_cubes_for_formulas — added in #643, yices.h:4748–4756
  - yices_generalize_model_with_budget / yices_generalize_model_array_with_budget — yices.h:4900–4904, Since 2.7.0

  Similarly, yices_new_model() (yices.h:3767, Since 2.6.4) currently appears only in a code example in model-operations.rst, not as a
  documented function entry.

  Missing error code in context-operations.rst

  Both yices_check_context_with_model and yices_check_context_with_model_and_hint document MCSAT_ERROR_ASSUMPTION_TERM_NOT_SUPPORTED
  in their error reports, but the header also lists MCSAT_ERROR_ASSUMPTION_TYPE_NOT_SUPPORTED (error 1002) as a possible return. The
  TYPE_NOT_SUPPORTED case should be added to both entries.
  
  Version annotation

  YICES_GEN_BY_PROJ_WIDE and the _with_budget variants are tagged "Since 2.7.0" in the headers. If the RST marks these as 2.8.0, that
  should be corrected to 2.7.0.

PR #645, opened 2026-06-28
Fix MCSAT function extensionality with diff witnesses (fixes 621)

### LINKED ISSUE #621: MCSAT: fix search-time extensionality for all function sorts (supersedes PR #590 guard)
This issue tracks replacing PR #590's temporary MCSAT guard with sound search-time support for equality and disequality over function sorts. The core bug is that MCSAT currently decides function equality by comparing its internal function ids rather than by extensional reasoning; this is unsound whenever the type forces an extensional equality that the Boolean abstraction can then contradict — concretely, finite-domain function sorts (where pigeonhole forces collisions) and unit-codomain function sorts (where all functions are equal).

A related secondary limitation affects model construction for self-referential or higher-order function values: during reconstruction, a recursive lookup for the same function id can observe a partially built function value (the fresh base, without the in-flight updates), yielding a model that is not extensional on the self-referential pair. The extensional-search plans above will also need to build extensional models, so this limitation is subsumed by the main fix and does not need a separate design.

Goals:
re-enable linux-4.0-rc1---kernel--locking--locktorture.ko.cil_smt-query.1.smt2.disabled
re-enable 32_1_cilled_ok_nondet_linux-3.4-32_1-drivers--input--keyboard--gpio_keys_polled.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out_smt-query.0.smt2.disabled
re-enable bool-array.smt2.disabled
delete wd_issue301.smt2.mcsat.gold so MCSAT answers match the shared sat gold
remove term_needs_function_diseq_guard from preprocessor.c
remove the KNOWN LIMITATION block in uf_plugin.c:875

disteph:
## Summary

Fixes #621 and supersedes the #590 MCSAT function-disequality guard by adding principled support for extensional function disequality in MCSAT.

This PR handles Yices function values across arbitrary arities, including higher-order function types, without tuple-encoding multi-argument applications or updates.

## Equality-Sensitive Types

A type is **equality-sensitive** when the current assertion/assumption context can make equality reasoning on values of that type semantically relevant.

The equality-sensitive set is computed as a least fixed point.

Seed rules:

- If an asserted equality atom or distinct predicate compares terms of type `T`, then `T` is equality-sensitive.
- If an asserted term has function type `A1 ... An -> B`, then each domain type `Ai` and the range type `B` are equality-sensitive.

Closure rule:

- If a tuple type `(A1, ..., An)` is equality-sensitive, then each component type `Ai` is equality-sensitive.
- If a function type `A1 ... An -> B` is equality-sensitive, then each domain type `Ai` and the range type `B` are equality-sensitive.

The set is context-dependent: it is scoped across push/pop, extended by temporary assumptions, and assumption-only sensitivity is cleared after solving with assumptions.

## Function-Type Policies

MCSAT now separates function reasoning by type policy.

- **Non-equality-sensitive function types:** function ids remain internal search values. MCSAT does not introduce function-value equality, function-value disequality, extensionality, or diff-witness obligations for these types. Weak-equality machinery may still track their updates and applications to enforce read-over-write semantics, but generated obligations are about the relevant index/result terms, not about equality of the function values themselves.

- **Equality-sensitive, non-risky function types:** function equality/disequality is semantically relevant, but no search-time diff witnesses are generated. Disequality commitments are satisfied during model construction using fresh values.

- **Equality-sensitive, risky function types:** function disequalities get search-time extensionality support using diff witnesses.

A function type is **risky** when its domain is finite, or when its range has unit cardinality. These are the cases where postponing witness construction can hide finite-domain conflicts or unavoidable equality conflicts from search.

## Invariants

The implementation enforces these invariants:

- generated function equality/disequality obligations only mention equality-sensitive types;
- equality sensitivity is scoped across push/pop and temporary assumptions;
- assumption-only sensitivity is cleared after solving with assumptions;
- type-risk results are context-independent and cached, while active equality-sensitive views are generation-checked;
- debug assertions reject generated equality obligations on non-sensitive types.

## Diff Witnesses

For an active disequality between same-typed functions

`f, g : A1 ... An -> B`

MCSAT creates witness terms

`d1 : A1, ..., dn : An`

and enforces the guarded extensionality condition

`f != g => f(d1, ..., dn) != g(d1, ..., dn)`.

The guard is required because the disequality source can be backtracked. Witnesses are cached persistently per ordered syntactic function pair, so the same pair does not generate fresh diff terms across branches.

If the witness applications themselves have function type, the same policy applies recursively. This terminates by structural decrease through the function range type.

Diff witness terms and their applications are mapped into produced models, so users can inspect the chosen witness values.

## Function-ID Semantics

For equality-sensitive function types, assigning distinct function ids is treated as a semantic disequality decision.

The implementation now:

- creates diff-witness obligations for distinct-id decisions on risky function types;
- records explicit non-risky function disequalities for model completion;
- keeps non-risky distinct ids satisfiable by assigning separated function values in the model;
- rejects incompatible equality merges between classes whose active function-id decisions cannot be reconciled;
- prevents cached function-id reuse from violating active disequality commitments.

## Weak Equality And N-Ary Updates

Weak-equality read-over-write reasoning is generalized from unary array-style updates to Yices n-ary function updates.

The implementation handles applications and updates componentwise, without introducing tuple terms.

Weak-equality reasoning is still used for read-over-write semantics even when the surrounding function type is not equality-sensitive. The equality-sensitivity gate only prevents weak equality from manufacturing function-level equality/extensionality obligations for types where function equality is not part of the current problem.

## Early Conflicts

For finite function types with exact cardinality, MCSAT can now detect disequality cliques that exceed the number of available function values.

Unit-range function disequality is also detected as an immediate conflict, including over infinite domains.

## Regression Coverage

This PR adds or updates coverage for:

- Bool-to-Bool function pigeonhole disequality;
- explicit and distinct-id function disequality;
- unit-range function disequality over infinite domains;
- higher-order finite-domain witnesses;
- higher-order non-risky disequality;
- non-risky model completion;
- incremental equality sensitivity;
- tuple sensitivity propagation;
- n-ary function update read-over-write, commutation, and permutation cases;
- finite-cardinality clique conflicts;
- prior Issue 414 and well-definedness regressions that previously hit the unsupported-theory guard.

## Testing

Local validation included:

- `make MODE=debug check-api`
- `make MODE=release check-api`
- Issue 621 UF regressions in debug and release modes
- focused Issue 414 / well-definedness regressions
- focused incremental array termination regression in debug and release modes

PR supported by Codex/GPT-5.5.


ahmed-irfan:
## Review

Reviewed the diff with a focus on the n-ary weak-equality generalization and its connection to the UF plugin. The core design holds up — I verified several alarming-looking paths are actually correct (see *Ruled out*). Findings most-severe first.

### Correctness

1. **`uf_plugin.c:2528-2529` — `null_value` flows unguarded into `vtbl_mk_update` (confirmed).** This PR removes both the preprocessor guard *and* the fail-closed `assert(false …unreachable…)` at `:2486`, so `make_fresh_function` can now legitimately return `null_value` at model build for higher-order / saturated finite function types (exercised by the new higher-order tests). Top-level callers null-check, but the recursive update loop in `get_function_value` does not: `result_value` (`:2528`) goes straight into `vtbl_mk_update` (`:2529`); `null_value` (−1) reaching `build_map_value` is dereferenced. Higher-order function over a small saturated range → ill-formed value / crash instead of a clean reject.

2. **`uf_plugin.c:1997` — `fun_diseq_trail_scan_index` not scoped across pop (confirmed).** `push`/`pop` (`:1957`/`:1975`) save only the entry-vector sizes; the scan cursor is just clamped to post-pop `trail_size`, and internal conflict backtracking doesn't bump the eqsens generation that resets it to 0. A function-disequality whose witness entry was created at a deeper level (freed on backtrack) and is then re-propagated to a *lower* physical index below the clamped cursor is never re-scanned → its extensionality witness is never regenerated → potential spurious SAT. Trigger needs lower-level propagation, so not trivial to reproduce, but the mechanism is real.

3. **`uf_plugin.c:2746` — shared function-id overwrite in `apply_search_diff_witnesses` (plausible).** Two active entries sharing a function id (e.g. `f≠g`, `f≠h`) each `vtbl_mk_update` the cached `function_value[f_id]`. If their diff terms evaluate to the same argument tuple and the witness apps are unvalued (per-pair range values guarantee only *intra*-pair distinctness), the second update can collapse the first disequality → non-extensional model. Worth an assertion / per-(id,point) guard.

### Cleanup / robustness

4. **`solver.c:660` — `eqsens_post_freeze_registration` is dead state** (set, never read). The intended tripwire for a post-freeze term introducing an uncovered function-equality type was never wired; in release the closure-completeness invariant has no enforcement (the debug assert is compiled out). Delete it and document why the closure is complete, or actually consume it.

5. **`weak_eq_graph.c` ↔ `uf_plugin.c` — ~220 lines of near-identical helpers duplicated** (`bool_term_is_true/false`, `terms_equal_in_branch`, `add_terms_eq[ual]_reason`, the sensitivity wrappers). A fix to one silently diverges from its twin — hoist into a shared util parameterized by ctx + eq_graph.

6. **`weak_eq_graph.c:1204` — weak path computed twice** in `weq_graph_array_weak_eq_at`: once with `path_cond = NULL`, then re-run identically into a throwaway `reason_edges` just to fill `path_cond` (`:1218`). Pass `path_cond` into the first call.

7. **`uf_plugin.c:1890` — redundant work in the hot path.** `check_fun_cardinality_conflict` runs twice per propagate; `find_fun_diseq_entry` / `active_diseq_blocks_function_id` are linear scans (`:328`/`:791`) inside O(active²) loops → O(n²–n³) as disequalities accumulate. A pair-keyed hash map (header already included) + a function-id→entries index removes the quadratic factor.

8. **`weak_eq_graph.c:30` — dead `USE_ARRAY_DIFF` unary path actively maintained.** Unreachable in every build, but refreshed onto the new helpers while keeping `assert(arity == 1)`. Subsumed by the n-ary `weq_graph_array_ext_check`. It's a trap: flipping the toggle yields a unary path whose arity-1 assert misfires on the n-ary terms this PR adds. Delete or gate behind a real build option.

Minor: `register_term → process_registration_queue → new_term_notify → register_term` recurses on the shared queue with depth bounded only by term nesting — likely fine, worth a glance.

### Ruled out (verified *not* problems)

- **INSTANCE_TYPE missing from the eqsens closure** — benign: the machinery is gated on the equality argument's own type being `FUNCTION_TYPE`; instance-typed terms are atomic UF, and standalone function terms are marked via the term-DAG walk regardless.
- **`equality_graph.c` congruence-over-pairs** — this is a *fix*: a congruence edge between pair nodes exists only when component reps collide, so components are in the same class and BFS-explainable for any node type; the old `EQ_NODE_TERM` asserts were never valid for nested pairs.
- **distinct-id cardinality-clique conflict literals** — valid: MCSAT conflict literals are evaluated by value, and the active-id view is rebuilt immediately before the check.
- **Witness `CAP` (16) → unsound SAT** — refuted: soundness rests on the cap-independent cardinality conflict + `make_fresh_function` distinctness. (Still: the caps at `:608`/`:1206` truncate *completeness* silently — a trace/stat when they bite would help.)

Findings #1 and #2 are the ones I'd block on — both are consequences of removing the preprocessor guard without fully covering the now-reachable paths.

disteph:
## Response to Review

This summarizes the changes made in response to the review of the PR, including correctness, cleanup, robustness, and items that were investigated but not changed.

## Correctness Findings

### 1. `null_value` could flow into `vtbl_mk_update`

**Review concern:**  
Removing the old unsupported-theory guard made higher-order and saturated finite function cases reachable. In those cases, `make_fresh_function()` can return `null_value`. Some model-building paths checked top-level results, but recursive update construction could pass `null_value` into `vtbl_mk_update()`, which would be unsafe.

**Action taken:**  
Model construction now guards these paths before building updates:

- `uf_model_builder_get_function_value()` checks whether `make_fresh_function()` returned `null_value` and stops constructing the function value if so.
- Recursive update construction checks argument values and result values before calling `vtbl_mk_update()`.
- Search diff witness application checks the function values, diff argument values, and witness application values before applying updates.
- Model disequality completion similarly checks that both function values are available before applying model-completion updates.

**Result:**  
The model builder now fails cleanly for unsupported/unconstructible function values instead of passing `null_value` into value-table update construction.

### 2. `fun_diseq_trail_scan_index` was not scoped

**Review concern:**  
The explicit function-disequality trail cursor was not saved/restored across push/pop. A witness entry created at a deeper level could be freed on backtrack, while the scan cursor remained past the relevant Boolean trail entry, causing the disequality to be missed later.

**Action taken:**  
`fun_diseq_trail_scan_index` is now part of the UF plugin scope state:

- It is saved in `uf_plugin_push()`.
- It is restored in `uf_plugin_pop()`.
- Disequality entries and model-disequality entries are also restored consistently with the same scope state.

**Result:**  
After backtracking, the explicit disequality scan cursor returns to the correct scope-local position, so reappearing/repropagated disequalities are not skipped.

### 3. Shared function-id updates could overwrite previous witness updates

**Review concern:**  
In `apply_search_diff_witnesses`, multiple active disequality entries can share a function id. Repeatedly updating the same cached function value could overwrite or collapse an earlier witness update if entries hit the same argument point.

**Action taken:**  
The model-building code was hardened around search diff witness application:

- Function values are fetched through the model builder before update application.
- Missing function values or missing argument/witness values cause the entry to be skipped rather than forcing an invalid update.
- Updates are written back to the function-value cache only after successful construction.

**Result:**  
The unsafe null/invalid-value overwrite path is guarded. The implementation does not add a separate per-`(function-id, point)` conflict table; the retained policy is to apply available witness updates defensively and avoid invalid model construction.

## Cleanup / Robustness Findings

### 4. Dead `eqsens_post_freeze_registration` state

**Review concern:**  
`eqsens_post_freeze_registration` was set/reset but never consumed, so it was dead state and did not enforce the intended invariant.

**Action taken:**  
Removed `eqsens_post_freeze_registration` from `mcsat_solver_t` and all initialization/reset paths.

Post-freeze registration is now explicitly documented: the pre-search freeze is the boundary for assertion/assumption roots, and terms generated during search do not enlarge the equality-sensitivity root set for the already-frozen generation.

**Result:**  
Dead state removed; the intended invariant is documented at the point where post-freeze registrations are ignored.

### 5. Duplicated Boolean branch helpers in UF and WEQ

**Review concern:**  
UF and weak-equality code duplicated Boolean literal/value checks. A future fix could easily update one copy but not the other.

**Action taken:**  
Added shared helpers:

- `src/mcsat/utils/branch_utils.c`
- `src/mcsat/utils/branch_utils.h`

UF and WEQ now both call:

- `mcsat_branch_bool_term_is_true()`
- `mcsat_branch_bool_term_is_false()`

The new source file is included in `src/Makefile`.

**Result:**  
The duplicated Boolean branch decoding logic is now centralized and the new files are tracked as build inputs.

### 6. WEQ weak path computed twice

**Review concern:**  
`weq_graph_array_weak_eq_at()` computed the weak path once to collect edges and then recomputed it to collect `path_cond`.

**Action taken:**  
Changed the first `weq_graph_compute_weak_path_at()` call to pass `path_cond` directly when the caller requests it, and removed the second computation plus temporary `reason_edges`.

**Result:**  
The same weak path is no longer recomputed just to collect reasons.

### 7. UF cardinality conflict check ran redundantly

**Review concern:**  
`uf_plugin_check_fun_cardinality_conflict()` ran twice per propagation in some paths, and other UF disequality lookups remain linear scans inside larger loops.

**Action taken:**  
Reduced duplicate cardinality checking in propagation:

- If an early cardinality check has already run after explicit witness additions and no new distinct-id witnesses were added afterward, the later check is skipped.
- If distinct-id witness generation adds new witness obligations, the later cardinality check still runs, because the relevant state has changed.

**Result:**  
The same cardinality scan is not repeated after unchanged witness state.

**Not done:**  
A pair-keyed hash map for `find_fun_diseq_entry()` and a function-id-to-entry index were not retained. A pair-map experiment was attempted, but it caused severe regressions on timeout-sensitive tests (`wd_issue294.smt2` and `aviad_true-alloca_true-termination.c.i.partial.smt2`). That change was reverted. The remaining vector scans are unchanged and should be treated as future optimization work, not part of this review response.

### 8. Dead unary `USE_ARRAY_DIFF` path

**Review concern:**  
The old WEQ `USE_ARRAY_DIFF` path was unreachable in normal builds, unary-only, and dangerous if toggled on after adding n-ary update support.

**Action taken:**  
Replaced the runtime `USE_ARRAY_DIFF` switch with a compile-time gate:

- `WEQ_ENABLE_LEGACY_ARRAY_DIFF`

The legacy diff fields, helpers, declarations, and dispatch are all behind this gate. Normal builds use the current n-ary extensionality path.

**Result:**  
The disabled unary path is no longer active dead code in the normal build and cannot be accidentally enabled by changing a local runtime constant.

### 9. Registration queue reentrancy

**Review concern:**  
`register_term -> process_registration_queue -> new_term_notify -> register_term` could recursively re-enter the shared registration queue drain, bounded only by term nesting.

**Action taken:**  
Added `registration_queue_processing` to `mcsat_solver_t`.

- If `register_term()` is called while the queue is already being drained, it enqueues/registers the term but does not recursively call `mcsat_process_registration_queue()`.
- The outer queue-drain loop continues until the queue is empty, so nested registrations are processed before the outer drain returns.

**Result:**  
Registration queue processing is now non-reentrant.

## Ruled-Out Review Items

The review also listed several items as “ruled out” rather than bugs. No code changes were made for these.

### `INSTANCE_TYPE` missing from equality-sensitivity closure

No change. The review concluded this was benign because function-equality sensitivity is driven by the equality argument’s function type and by walking standalone function terms in the term DAG.

### Equality graph congruence over pair nodes

No change. The review concluded the change was valid: congruence edges between pair nodes are explainable through component representatives, and the old term-node-only asserts were too strong.

### Distinct-id cardinality clique conflict literals

No change. The review concluded these literals are valid in MCSAT because conflicts are evaluated by value and the active-id view is rebuilt immediately before the check.

### Witness cap causing unsound SAT

No semantic change. The review concluded soundness does not rely on the witness cap because finite-cardinality conflicts and model construction provide the cap-independent coverage.

**Not done:**  
No new trace/stat was added for witness-cap truncation. That remains a possible diagnostics improvement, not a correctness fix.

## Validation

Local validation after the review response:

- `make MODE=debug`
- `git diff --check`
- `git diff --cached --check`
- `tests/regress/both/wd_issue294.smt2`
- `tests/regress/mcsat/arrays/aviad_true-alloca_true-termination.c.i.partial.smt2`
- `tests/regress/both/issue414_bool_bool_pairwise_diseq.smt2`

Observed local timings for the timeout-sensitive Aviad test:

- Baseline before cleanup changes: about `24.980s`
- Final run after review response: about `22.480s`

No plan or review markdown files are part of the code changes.

ahmed-irfan:
A few redundant bits in the current diff:

1. **Duplicated equality-sensitivity wrappers.** `uf_type_is_equality_sensitive` (`uf_plugin.c:219`) ≡ `weq_graph_type_is_equality_sensitive` (`weak_eq_graph.c:112`), and `uf_plugin_equality_sensitivity_is_frozen` (`uf:256`) ≡ `weq_graph_equality_sensitivity_is_frozen` (`weq:122`) — identical `ctx`-callback one-liners. `branch_utils.{c,h}` already holds this kind of shared helper; fold these in too. (The clearest one to fix.)

2. **Thin bool wrappers.** `uf_plugin_bool_term_is_true/false` (`uf:967/971`) and the weq equivalents are now one-line passthroughs to `mcsat_branch_bool_term_is_*`; call sites could use the shared fn directly.

3. **Hard tabs** in a 2-space-indented file at `uf_plugin.c:1878–1879, 2740, 2761, 2974` — whitespace only.

4. **`add_terms_eq_reason` (weq) / `add_terms_equal_reason` (uf)** duplicate the eq-graph-value reason logic — only partly, since the uf version also validates each reason literal is true-in-branch.

5. **Legacy `WEQ_ENABLE_LEGACY_ARRAY_DIFF` path** (~176 lines) is compiled out in all shipped builds and subsumed by the n-ary `weq_graph_array_ext_check` — kept as reference behind the flag.

#1 is the only one really worth folding; #3 is trivial.

ahmed-irfan (approved):
Reviewed the latest revision, focused on the new function-id representative mechanism (`5fe4e3c`, `d517034`) on top of the n-ary weak-equality / equality-sensitivity work.

The eq-graph now tracks a representative function term per function-id value node, and the active-function-id view is driven off those reps instead of a trail scan. I traced the soundness-critical paths and found no correctness issues:

- **Id-reuse is sound** — the dynamic class-traversal compat check (`uf_plugin_function_id_is_term_compatible`) walks the full value-class ring and blocks reuse whenever a member is provably `disequal_terms` under `compatible_types`; value nodes are unique per id and every function carrying an id is merged into that class, so no conflicting member hides outside it.
- **Removing the rebuild-level `diff_witness_terms` / search-diff filters is safe** — both are enforced at point-of-use (the `add_fun_diseq_pair` backstop, distinct-id witness gen, and the cardinality check), and the one unfiltered consumer (id reuse) is a benign heuristic.
- **The `continue`→`assert` conversions hold** — the invariants are guaranteed by the record guards + lock-step push/pop undo, so release builds can't push a bad entry.
- **Permanent→scoped id/type binding** — not a regression; nothing downstream needs a global id→type map, and pop restores the scoped state.

CI is green across the matrix (incl. debug/gcov, thread-safety, Windows), coverage +0.5%, and the new targeted regressions cover each path. LGTM.

Optional, non-blocking cleanups:
1. `eq_graph_value_class_has_function_term` / `eq_graph_function_value_rep_has_trail_edge` are public but assert-only — under `NDEBUG` they're external-linkage functions with zero callers. Consider `#ifndef NDEBUG` per the convention at `equality_graph.c:165-189`.
2. The inner assert branch in `eq_graph_propagate` (~`:1350`) is unreachable (a function term in a value class always has a non-NULL rep) — delete or comment as a can't-happen invariant.
3. Minor: `eq_graph_value_id_if_exists` overlaps `eq_graph_value_id`/`eq_graph_has_value` (unnecessary const-cast on an already-const arg); redundant `assert` after the sensitivity early-return in `uf_plugin_add_fun_diseq_pair` (~`:620`).

PR #656, opened 2026-07-20
Fix signed-overflow UB in bv64 interval abstraction

ahmed-irfan:
Some QF_BV benchmarks returned `unsat` instead of `sat` on Linux/GCC release builds (macOS/clang was unaffected). Not delegate-specific — the default solver is equally wrong.

**Root cause:** `bv64_interval_abstraction.c` computed `x+y` / `x-y` and negations as signed `int64_t` (undefined behavior on overflow), then detected the overflow by inspecting the result's sign. At `-O2/-O3` GCC assumes signed overflow cannot happen and deletes those checks, so overflow goes undetected → unsound bitvector interval → wrong `unsat`.

**Fix:** perform the wraparound arithmetic and negations in the unsigned domain (well-defined, identical two's-complement result) so the overflow checks survive optimization. Covers `sum`/`diff`/`opposite`/`mul`/`negate`/`power`.

**Tests:** two QF_BV regression tests that return `unsat` on the buggy build and `sat` with the fix. Full regression suite passes.


disteph (approved):
Review supported by Codex/GPT5.6Sol after additional commit fc17944

# Code Review: SRI-CSL/yices2 PR #656

**PR:** [Fix signed-overflow UB in bv64 interval abstraction](https://github.com/SRI-CSL/yices2/pull/656)  
**Review date:** 2026-07-24  
**Base:** [`84efa9276793e33281cecaf3271215151b1947ac`](https://github.com/SRI-CSL/yices2/commit/84efa9276793e33281cecaf3271215151b1947ac)  
**Head:** [`fc179443932cc6cac85c618b0847e5800f5889f9`](https://github.com/SRI-CSL/yices2/commit/fc179443932cc6cac85c618b0847e5800f5889f9)  
**Commits reviewed:** 2  
**Change size:** 8 files, 928 insertions, 1187 deletions

## Recommendation

**Approve.**

The production fix is correct for Yices's supported two's-complement GCC/Clang
targets. I found no remaining signed-overflow defect in the modified arithmetic,
and the fix restores correct results on every reported SMT-COMP benchmark tested.

I found no correctness, regression, or test-integration issue that should block
or qualify approval. The two added SMT regressions are discovered by `make
check`, and CI invokes that target.

## Correctness Assessment

The root-cause analysis in the PR is accurate. The old code performed signed
`int64_t` arithmetic and then tried to detect overflow by inspecting the wrapped
result. Signed overflow is undefined in C, so an optimizing compiler may assume
that it never occurs and remove the checks. That can produce an interval that is
too narrow, which is unsound for an abstraction used by the solver.

The replacement consistently moves the potentially overflowing operation into
the unsigned domain, where arithmetic wraps modulo \(2^{64}\):

| Area | Change | Assessment |
| --- | --- | --- |
| `opposite` | Cast before unary negation | Correctly handles `INT64_MIN` without signed UB. |
| `sum` | Convert both operands to `uint64_t` before addition | Preserves the wrapped bit pattern and keeps the 64-bit sign-based overflow test valid under optimization. |
| `diff` | Convert both operands to `uint64_t` before subtraction | Same reasoning as `sum`; the existing overflow predicate remains valid. |
| `mul` absolute values | Negate unsigned representations | Correctly computes the magnitude of `INT64_MIN` as \(2^{63}\). |
| `mul` negative result | Negate the unsigned magnitude | Avoids `-INT64_MIN` in the signed domain. |
| interval negation | Negate `a->high` as unsigned | Removes the remaining direct `INT64_MIN` negation. |
| even powers | Compute `abs(a->low)` as unsigned | Correctly compares the magnitudes when the lower bound is `INT64_MIN`. |

For widths below 64, the mathematical results used by `sum` and `diff` remain
representable in `int64_t`; the existing `fits_k_bits` checks correctly detect
whether the result needs another bit. At width 64, the sign predicates detect
the two signed-overflow cases after unsigned wraparound.

I also inspected the rest of `bv64_interval_abstraction.c` for the same class of
operation. The remaining signed negation in `num_significant_bits` starts from
`INT64_MIN / 2` and is representable. The additions used to construct bounds
are constrained to representable ranges. I found no missed instance of the
reported undefined behavior.

### Portability note

Converting an out-of-range `uint64_t` value to `int64_t` is
implementation-defined in pre-C23 C. The patch relies on GCC/Clang's
two's-complement conversion, as the surrounding Yices code already does and
explicitly documents. This is appropriate for the supported competition and CI
targets and is not a regression introduced by this PR. It would need a separate
design decision only if Yices intends to support non-two's-complement or
trapping conversions.

## Test Assessment

The two added SMT files are exact copies of reported SMT-COMP failures:

- `bv64_interval_signed_overflow_a.smt2` is `predicate_484.smt2`;
- `bv64_interval_signed_overflow_b.smt2` is `predicate_1247.smt2`.

Both have `sat` gold files and fail with `unsat` on the affected binary. They are
good end-to-end regressions because they exercise the actual simplification and
solver path that produced the competition error. They live under
`tests/regress/bv`, so the repository's `make check` target discovers and runs
them through `tests/regress/check.sh`. CI calls `make check` before `make test`.

Under the existing harness, `tests/unit/*.c` programs are built uniformly by
`make test`, while runnable solver regressions are checked by `make check`.
Replacing the two old unit programs is consistent with that organization and is
reasonable:

- `test_int64.c` copied private arithmetic helpers into the test itself, so it
  tested duplicate code rather than the production implementation.
- `test_bv64_abstraction2.c` exercised production functions over many inputs
  but only printed results; it had no oracle and no failure checks for result
  correctness.
- `test_bv64_interval_boundaries.c` calls production functions and has explicit
  expected abstractions, making it a better focused validation program within
  the existing unit-test harness.

The new boundary cases cover the important signed endpoints in both operand
orders where appropriate. Their expected use of the top abstraction on signed
overflow is conservative and sound.

## Verification Performed

### Local environment

- Linux x86-64;
- GCC 11.4.0;
- release mode (`-O3 -fomit-frame-pointer -DNDEBUG`);
- no MCSAT or CaDiCaL dependency enabled in the local review build.

### Results

| Check | Result |
| --- | --- |
| Release build at PR head | Passed |
| `git diff --check` | Passed |
| `make check` release regression suite | 1030 passed, 0 failed |
| `make test` unit-program build | Passed |
| Added SMT regressions | Both returned `sat` |
| New boundary unit executable, run directly | Passed |
| ASan/UBSan `-O3` build of unit tests | Passed |
| New boundary unit executable under ASan/UBSan | Passed with no diagnostics |
| Current GitHub checks on PR head | 30 successful, 2 skipped, 0 failed |

Nine CaDiCaL-only regression files were skipped by the local build because it
did not include CaDiCaL. This patch is independent of the delegate, and the
repository's current GitHub matrix, including MCSAT and thread-safe
configurations, is green.

The fixed release binary also returned `sat` on all five QF_BV benchmarks named
by the SMT-COMP organizers:

| Benchmark | Result |
| --- | --- |
| `readelf/predicate_1247.smt2` | `sat` |
| `readelf/predicate_484.smt2` | `sat` |
| `readelf/predicate_494.smt2` | `sat` |
| `readelf/predicate_7707.smt2` | `sat` |
| `symbolic_memory/linear/hdp/predicate_2973.smt2` | `sat` |

## Final Assessment

The arithmetic patch fixes the diagnosed unsoundness without changing behavior
for non-overflowing inputs. The source change is narrowly scoped, internally
consistent, and validated by both end-to-end and direct boundary tests. I found
no review findings. The PR is ready to merge.

PR #658, opened 2026-07-22
Fix ef-solve crash and spurious "model error" on UF-in-arithmetic quantifiers (#657)

### LINKED ISSUE #657: FATAL ERROR: ef-solve failed: unexpected status: model error
The following `.ys` file will crash Yices-2.7.0:

```
; test.ys
(define cryWidthUnknown::(-> int int))
(define k_2_to_64::int 18446744073709551616)
(assert (forall (x::int) (or (> (cryWidthUnknown x) 64) (< x k_2_to_64))))
(ef-solve)
```
```
$ yices --version
Yices 2.7.0
Copyright SRI International.
Linked with GMP 6.3.0
Copyright Free Software Foundation, Inc.
Build date: 2026-06-25
Platform: x86_64-pc-linux-gnu (release/static)

$ yices --mode=ef test.ys

*************************************************************
FATAL ERROR: ef-solve failed: unexpected status: model error

*************************************************************
Please report this bug to yices-bugs@csl.sri.com.
To help us diagnose this problem, please include the
following information in your bug report:

  Yices version: 2.7.0
  Build date: 2026-06-25
  Platform: x86_64-pc-linux-gnu (release/static)

Thank you for your help.
*************************************************************
```

ahmed-irfan:
Fixes #657. `ef-solve` on a forall body with an uninterpreted-function application inside arithmetic (e.g. `(f x)` in `(< x C)`) crashed, and on release builds aborted with a bogus `FATAL ERROR: ... model error` bug report. Two independent causes:

- **Crash (UB):** `quant_pattern.c` walked terms with `term_child`, which only supports composite terms. Arithmetic/bitvector polynomials, products, and projections (still counted by `term_num_children`) hit `term_child`'s composite path and read a wrong descriptor. Added a shared `term_ith_subterm()` in `term_explorer.c` that reads those components off the descriptor, and use it in the three quant pattern walkers.
- **Spurious bug report:** `ef_generalize3` builds a full model via `yices_model_from_map`, which rejects function-typed existentials. A problem with an uninterpreted-function existential and arithmetic universals is unsupported by projection-based generalization, so it now returns `unknown` instead of a fatal bug.

Adds regression test `tests/regress/efsmt/ef-tests/iss657.ys`. All 108 efsmt tests pass on release/debug/sanitize.


disteph (approved):
# Code Review (supported by Codex/GPT5.6Sol: SRI-CSL/yices2 PR #658

**PR:** [Fix ef-solve crash and spurious "model error" on UF-in-arithmetic quantifiers (#657)](https://github.com/SRI-CSL/yices2/pull/658)  
**Review date:** 2026-07-25  
**Base:** [`6435defcfce65f5a262ad7766ee963ac22532da2`](https://github.com/SRI-CSL/yices2/commit/6435defcfce65f5a262ad7766ee963ac22532da2)  
**Head:** [`ed5fbafb62694a7f275c34718c163c45daa8b2b8`](https://github.com/SRI-CSL/yices2/commit/ed5fbafb62694a7f275c34718c163c45daa8b2b8)  
**Commits reviewed:** 1  
**Change size:** 7 files, 112 insertions, 15 deletions

## Findings

**No findings.**

I found no correctness, soundness, memory-safety, or regression issue in the
current PR diff.

## Recommendation

**Approve.**

The patch fixes both reported failure modes without claiming support for a
theory combination that the projection-based EF solver cannot handle. The
generic traversal change is consistent with Yices's internal term
representations, and the new regression is run by `make check`.

## Correctness Assessment

### Generic term traversal

The original quantifier-pattern walkers used `term_num_children` followed by
`term_child`. Those APIs did not have matching domains: `term_num_children`
counts components of polynomial, power-product, and projection descriptors,
whereas `term_child` requires a composite term. Passing one of the former term
kinds to `term_child` can therefore assert in a debug build or read the wrong
descriptor in a release build.

The new
[`term_ith_subterm`](https://github.com/SRI-CSL/yices2/blob/ed5fbafb62694a7f275c34718c163c45daa8b2b8/src/terms/term_explorer.c#L493)
closes that API mismatch:

| Term kind | Access used | Assessment |
| --- | --- | --- |
| `ARITH_POLY` | `poly_term_desc(...)->mono[i].var` | Correct; a constant monomial is represented by `const_idx` and is skipped. |
| `ARITH_FF_POLY` | `finitefield_poly_term_desc(...)->mono[i].var` | Correct for the finite-field polynomial descriptor. |
| `BV64_POLY` | `bvpoly64_term_desc(...)->mono[i].var` | Correct for small bitvector polynomials. |
| `BV_POLY` | `bvpoly_term_desc(...)->mono[i].var` | Correct for general bitvector polynomials. |
| `POWER_PRODUCT` | `pprod_term_desc(...)->prod[i].var` | Correct; exponents are metadata, not subterms. |
| `SELECT_TERM`, `BIT_TERM` | `proj_term_arg(...)` | Correct; both are unary projection descriptors. |
| Negative or composite terms | Existing `term_child` | Preserves the established behavior for supported composite terms. |

The precondition uses `good_term` rather than `term_is_composite`, which is
necessary for the expanded domain, and still checks the index against
`term_num_children`. Returning `NULL_TERM` for a constant monomial is explicit
in the header contract and handled by all three changed callers.

The changes in
[`quant_pattern.c`](https://github.com/SRI-CSL/yices2/blob/ed5fbafb62694a7f275c34718c163c45daa8b2b8/src/solvers/quant/quant_pattern.c)
use the new accessor consistently in pattern processing, single-pattern
inference, and multi-pattern inference. Interpreted operators are traversed to
find nested variables and applications, but are not themselves accepted as
application patterns. Replacing the previous default assertions with
conservative `break`/`skip` behavior is appropriate for these newly reachable
term kinds.

### EF generalization status

`ef_generalize3` constructs a model from the combined existential and universal
assignment. The underlying `map_to_model` interface explicitly does not support
function types, so a function-typed existential makes this projection strategy
unavailable.

The new
[`ef_cnstr_has_function_evar`](https://github.com/SRI-CSL/yices2/blob/ed5fbafb62694a7f275c34718c163c45daa8b2b8/src/exists_forall/efsolver.c#L993)
checks exactly the existential variables included in that constraint's full
map. If model construction fails in this known unsupported case,
[`ef_generalize3`](https://github.com/SRI-CSL/yices2/blob/ed5fbafb62694a7f275c34718c163c45daa8b2b8/src/exists_forall/efsolver.c#L1014)
sets `EF_STATUS_UNKNOWN`. Other model-construction failures retain the existing
`EF_STATUS_MDL_ERROR` path and error code. This avoids the spurious fatal bug
report without converting an unsupported problem into an unsound `sat` or
`unsat` result.

The early return does not leak a model: `yices_model_from_map` returned `NULL`,
and any previous `full_model` was freed before model construction.

## Test Assessment

The added regression
[`iss657.ys`](https://github.com/SRI-CSL/yices2/blob/ed5fbafb62694a7f275c34718c163c45daa8b2b8/tests/regress/efsmt/ef-tests/iss657.ys)
contains the essential combination:

- an existential uninterpreted function;
- a universal integer variable;
- the function application nested in an arithmetic formula;
- `ef-solve` in EF mode.

Its gold result is `unknown`, which verifies both externally visible repairs:
the traversal no longer crashes and the unsupported model construction no
longer becomes a fatal model error. The `.options` file selects `--mode=ef`,
and the regression harness discovers the test through `make check`.

The test is intentionally end-to-end. It exercises the arithmetic-polynomial
path that caused issue #657 and the subsequent status handling. It does not
individually instantiate every descriptor supported by `term_ith_subterm`, but
the implementation for those descriptors is direct and mirrors
`term_num_children`; I do not consider that a merge-blocking coverage gap.

## Verification Performed

### Local environment

- Linux x86-64;
- GCC 11.4.0;
- GMP enabled;
- MCSAT and external SAT delegates disabled in the local build.

### Results

| Check | Result |
| --- | --- |
| `git diff --check` on current base-to-head diff | Passed |
| Release build (`-O3 -DNDEBUG`) | Passed |
| Full release `make check` | 1,031 passed, 0 failed |
| New regression in release mode | `unknown` |
| Debug build with assertions enabled | Passed |
| New regression in debug mode | `unknown`, no assertion failure |

## Final Assessment

The patch repairs an invalid internal traversal, preserves conservative pattern
inference for interpreted terms, and reports the unsupported EF case as
`unknown` rather than aborting. The implementation is narrowly scoped and the
regression reproduces the complete user-visible path. I found no review
findings and recommend approval.

PR #664, opened 2026-09-16
Explicitely set `-std=gnu11`

Ovascos:
Currently yices does not set a C standard and uses the compiler's default standard. This PR adds `-std=gnu11` to the compiler arguments.
It should be sufficient to compile with `gnu99`. To be on the safe side and future proof, we use `gnu11`, which is supported in any modern (and even not so modern) version of gcc and clang. This PR also updates the documentation and notes gcc 4.9 and clang 3.3 as the minimal required versions. Those are 10+ years only and should be widely available.
The configure script checks the compiler versions and verifies that the compiler supports `-std=gnu99` as well as `-Wpedantic` which is already required by some parts of the code.


src/Makefile
```diff
@@ -780,6 +780,7 @@ ifeq ($(POSIXOS),mingw)
 endif
 
 
+CFLAGS += $(C_STD)
```
ahmed-irfan: **Blocker: the Debian build gets no `-std=gnu11`.**

`+=` does not apply when CFLAGS comes from the command line. GNU make drops a makefile assignment to a command-line variable unless you write `override`.

`debian/rules:12` and `debian/rules:15` run `$(MAKE) CFLAGS="$(CFLAGS)" ...`. The Debian build therefore compiles with the default standard of `$(CC)`.

`debian/rules` also uses `dh --with autoreconf`. Autoconf 2.72 and newer put `-std=gnu23` into `CC`. I confirmed `CC=gcc -std=gnu23` in a generated `make.include` with autoconf 2.73. Debian would ship a C23 build while every other build uses C11. That is the split this PR must prevent.

I verified the make behavior with a minimal makefile:

```
$ make -f t.mk
CFLAGS=[ -std=gnu11]
$ make -f t.mk CFLAGS=-O2
CFLAGS=[-O2]
```

Use `override CFLAGS += $(C_STD)`, or put `$(C_STD)` in the compile and link recipes. `tests/api/Makefile:112` and `tests/unit/Makefile:112` need the same fix.

configure.ac
```diff
@@ -117,7 +117,54 @@ else
...
+dnl
+dnl Check that the C compiler accepts -Wpedantic
+dnl --------------------------------------------
+dnl
+dnl The API tests (tests/api) are compiled with -Wpedantic. GCC rejects an
+dnl unknown warning option, but Clang only warns about it, so the check adds
+dnl -Werror. Configure fails if the flag is not accepted.
+dnl
+csl_saved_cflags=$CFLAGS
+CFLAGS="$C_STD -Wpedantic -Werror $CFLAGS"
```
ahmed-irfan: **Blocker: this probe aborts configure for a common macOS build.**

```
$ ./configure CFLAGS="-L/opt/homebrew/lib"
checking whether gcc -std=gnu23 accepts -Wpedantic... no
configure: error: *** gcc -std=gnu23 does not accept -Wpedantic, which Yices requires ***
```

Clang warns `-Wunused-command-line-argument` for `-L` during a `-c` compile. The `-Werror` on this line promotes that warning to an error. The message then blames the compiler for a flag the compiler accepts. `src/Makefile:704` already adds `-Qunused-arguments` on darwin for this warning class.

The probe body is an empty program. A user flag can only make the probe fail for a false reason. Drop `$CFLAGS` from the probe:

```
CFLAGS="$C_STD -Wpedantic -Werror"
```

configure.ac
```diff
@@ -117,7 +117,54 @@ else
...
+dnl POSIX/BSD interfaces the sources use (strdup, strerror_r, fdopen, fileno,
+dnl strtok_r, getc_unlocked, getpagesize, pthread_mutexattr_settype, ...).
+dnl
+dnl We check two things: that $CC accepts the flag, and that it actually
+dnl delivers C11. A compiler that accepts -std=gnu11 but reports an older
+dnl __STDC_VERSION__ is rejected. Configure fails if either check fails.
+dnl
+C_STD="-std=gnu11"
+csl_saved_cflags=$CFLAGS
+CFLAGS="$C_STD $CFLAGS"
```
ahmed-irfan: **Blocker: the probe uses a flag order that the build never uses.**

This line puts `$C_STD` first. `src/Makefile:783` appends `$(C_STD)` last. A user `-std=` therefore wins the probe and loses the build.

```
$ ./configure CFLAGS="-std=c99"
checking whether gcc supports -std=gnu11... no
configure: error: *** gcc does not support -std=gnu11. Yices requires a C11 compiler: GCC 4.9 or newer ***
```

The compiler does support gnu11. The probe compiled `gcc -std=gnu11 -std=c99`, so `__STDC_VERSION__` was 199901L. The build would have used gnu11.

Use `CFLAGS="$CFLAGS $C_STD"` to match the Makefiles.

ahmed-irfan (commented):
Review of the `-std=gnu11` change. Three blocking findings follow as inline comments. I built the PR head in a throwaway worktree with autoconf 2.73 and Apple clang, and I reproduced all three.

src/Makefile
ahmed-irfan: Fixed. `override CFLAGS += $(C_STD)` is now in `src/Makefile`, `tests/api/Makefile` and `tests/unit/Makefile`. `make CFLAGS="-O2 -g -L/opt/homebrew/lib"` from the top level ends the compile line with `-std=gnu11 -fPIC -c frontend/yices.c`. The placement comment is correct. A plain `CFLAGS := -O3` after the `override` line disappears.

configure.ac
ahmed-irfan: Fixed. The probe now uses `CFLAGS="$C_STD -Wpedantic -Werror"`. `./configure CFLAGS="-L/opt/homebrew/lib"` exits 0, and both probes report `yes`.

configure.ac
ahmed-irfan: Fixed. The probe now uses `CFLAGS="$CFLAGS $C_STD"`. `./configure CFLAGS="-std=c99"` reports `yes` for both probes. `make show-details` in `src` then reports `CFLAGS = -std=c99 ... -std=gnu11`.

ahmed-irfan (approved):
All three blockers are fixed. I rebuilt the PR head and reproduced each case. Environment: macOS arm64, Apple clang as `gcc`, autoconf 2.73, GNU make 3.81.

- `make CFLAGS="-O2 -g -L/opt/homebrew/lib"` from the top level ends the compile line with `-std=gnu11`. The build exited 0.
- `./configure CFLAGS="-L/opt/homebrew/lib"` exits 0. Both probes report `yes`.
- `./configure CFLAGS="-std=c99"` reports `yes` for both probes. `make show-details` in `src` then reports `CFLAGS = -std=c99 ... -std=gnu11`.
- `make check-api` compiles the three exempt files with `-std=gnu11` and without `-Wpedantic -Wextra`. The global `override` does not block the per-target `CFLAGS :=`. I checked make 3.81 only. The test binaries did not run here. CI ran them, and CI is green.
- The placement comment is correct. A plain `CFLAGS := -O3` after the `override` line disappears.

Nit: the PR description says that configure checks `-std=gnu99`. The code checks `-std=gnu11`.

Out of scope, and older than this PR: a command-line `CFLAGS=` still discards the plain `CFLAGS :=` additions, so the Debian shared library likely loses `-fvisibility=hidden`. I did not check with `nm`.

Ovascos:
There are further parts of the makefiles that are long outdated. Maybe we should consider switching to `cmake` eventually.

